PR #3, opened 2019-09-06
Inline error styling

roblevintennis:
https://app.asana.com/0/1136962714401929/1138726622315604/f

image


src/components/InfoMessage/InfoMessage.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Body } from '../index'
+import { Footnote } from '../index'
```
roblevintennis: Better matches Ryan's specs

PR #14, opened 2019-09-18
[WIP] Changes related to quote branch

zach-ethos:
Don't merge; just want to get a diff and thread for recent changes in this branch.


zach-ethos:
Having looked at the diff, I agree we don't want to merge this into master.

PR #15, opened 2019-09-20
Add type='tel' to ZipInput

zach-ethos:


src/components/Inputs/ZipInput/ZipInput.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { TextInput } from '../TextInput'
 
 export const ZipInput = (props) => {
-  return <TextInput {...props} />
+  return <TextInput {...props} type="tel" />
```
bwalker8038: Using `inputmode=numeric` will work for [Modern-ish browsers](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) (Chrome 66+ on Android and Safari (Webkit) 12.2+ on iOS). This also avoids Android displaying the telephone numeric pad (which is slightly different than the default numeric pad).

See: https://css-tricks.com/html-for-zip-codes/

zach-ethos: Huh. The browser support.... actually looks horrific for a sec until you remember we just need it on mobile :)

PR #16, opened 2019-09-23
[WIP] Use numeric keypad on mobile in Zip and Birthdate inputs

zach-ethos:
Succeeds #15.


src/components/Inputs/BirthdateInput/BirthdateInput.test.js
```diff
@@ -1,6 +1,6 @@
-import React from 'react';
-import { BirthdateInput } from './BirthdateInput';
-import renderer from 'react-test-renderer';
```
roblevintennis: 🤦‍♂ this is going to be a very hard habit to break lol. I think we gotta get eslint over into DS lol

src/components/Inputs/BirthdateInput/__snapshots__/BirthdateInput.test.js.snap
```diff
@@ -37,6 +38,7 @@ Array [
     }
     name="birthdate-auto-corrected"
     onBlur={[Function]}
+    pattern="[0-9]*"
```
roblevintennis: Can you remind me what's the '*' doing on this particular pattern?

src/components/Inputs/BirthdateInput/__snapshots__/BirthdateInput.test.js.snap
```diff
@@ -20,6 +20,7 @@ Array [
     className="BirthdateInput TextInput"
     data-tid="le-birthdate"
     guide={true}
+    inputMode="numeric"
```
roblevintennis: Is this approach working for us in terms of when you spot check?

src/constants.js
```diff
@@ -0,0 +1,6 @@
+// https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode
+export const INPUT_MODES = { NUMERIC: 'numeric' }
```
roblevintennis: Are we anticipating some more input modes? These always pretty much map to a value for [inputmode attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) right?

The reason I ask above is I'm wondering if we should flatten this out if we don't really plan to use inputmodes for much else then kicking in numeric on mobile. For example, maybe we just make this `export const INPUT_NUMERIC = 'numeric'` so other devs don't have to mentally parse as much. Maybe same for the pattern. Thoughts?

src/components/Inputs/BirthdateInput/BirthdateInput.js
```diff
@@ -99,6 +101,8 @@ const PrivateBirthdateInput = (props) => {
           !!err ? 'BirthdateInput TextInput Error' : 'BirthdateInput TextInput'
         }
         type="text"
+        inputMode={INPUT_MODES.NUMERIC}
```
roblevintennis: I really like how you leveraged DI on zip and birthdate here @zach-ethos 👍 nice job! I think it's flexible but still super simple in terms of understandability

src/components/Inputs/TextInput/TextInput.js
```diff
@@ -141,6 +148,11 @@ function PrivateTextInput({
 PrivateTextInput.PUBLIC_PROPS = {
   'data-tid': PropTypes.string.isRequired,
   disabled: PropTypes.bool,
+  // This is a newer way to trigger the mobile numeric keypad (old way was
+  // `type='tel'`.) For now, we probably either want to leave this unset, in
+  // which case it is equivalent to 'text', or set 'numeric' for numeric inputs.
+  inputMode: PropTypes.oneOf(Object.values(INPUT_MODES)),
```
roblevintennis: I doubt this broke anything, but did you happen to spot check zip, bd, and textinput in the styleguidist based DS?

src/components/Inputs/ZipInput/ZipInput.js
```diff
@@ -1,9 +1,16 @@
...
+import { INPUT_MODES, INPUT_PATTERNS } from '../../../constants'
 import { TextInput } from '../TextInput'
 
 export const ZipInput = (props) => {
-  return <TextInput {...props} />
+  return (
+    <TextInput
+      {...props}
+      inputMode={INPUT_MODES.NUMERIC}
+      pattern={INPUT_PATTERNS.NUMERIC}
```
roblevintennis: So this is probably only needed for envs NOT mobile huh? I'm thinking the inputmode on mobile will give the numeric UI that only allows 0-9 anyhoo huh? Just thinking aloud. Yeah, but we still need for desktop et al huh?

src/components/Inputs/BirthdateInput/__snapshots__/BirthdateInput.test.js.snap
bwalker8038: `[0-9]*` should match any `n` amount of characters contained in the the `[0=9]` set, where `n >= 0`

roblevintennis: Ok, so it's a literal regex metacharacter thx for confirming @bwalker8038

src/components/Inputs/BirthdateInput/BirthdateInput.test.js
zach-ethos: It's helpful IMHO (if you're using Sublime) to set up the Prettier plugin so you can format on save or with a key combo. https://github.com/getethos/ethos/blob/trunk/devops/dev/Sublime%20Text%203/JsPrettier.sublime-settings

src/components/Inputs/BirthdateInput/BirthdateInput.js
zach-ethos: Can you explain what you mean by DI here? (IRL maybe)

src/components/Inputs/BirthdateInput/__snapshots__/BirthdateInput.test.js.snap
zach-ethos: I think not having this means the regular keyboard is open, but with the modifier that opens the number keys on the top row (instead of the 3x3 pad we want).

zach-ethos:
Closing till we sort out bigger problems and I have a chance to confirm this works.

PR #19, opened 2019-09-25
Expose restrict as helper

roblevintennis:
[Asana](https://app.asana.com/0/1137984670709817/1141575727614402/f)

We need to share the restrict functionality so that other code doesn't duplicate it:
https://github.com/getethos/ethos/pull/1139

This task updates the design system to move it out to its own helper, then update TextInput to use the helper (currently that's the only place we do this in DS).

Subsequently, we will add design system 1.1. to the monorepo package json and the above referenced PR's code will be able to pull that in and remove the duplication.
----

Unfortunately, I have to remove unrelated onChange: PropTypes.func code because the usage is throwing errors when I check the TextInput component in styleguide:

image


src/components/Inputs/TextInput/TextInput.js
```diff
@@ -129,7 +123,6 @@ PrivateTextInput.PUBLIC_PROPS = {
   labelCopy: PropTypes.string.isRequired,
   validator: PropTypes.func,
   onBlur: PropTypes.func,
-  onChange: PropTypes.func,
```
roblevintennis: I'm bummed about this. I suppose I could have left this and cut a branch off `master`, but since my changes directly affect this component I'm not able to spot check with full confidence without removing the props.onChange.

But now I have to worry that I've somehow caused a regression downstream. The intent of this whole PR should be risk free: expose the `restrict` method as a helper, make sure TextInput still works, pull it into monorepo and use the helper.

So, above said, since monorepo is using v1.0 (legacy DS) and doesn't even reference any of this code yet, and, CMS is pointed to the "quote crazy" branch for the time being, I suppose there's no risk 🤔 😞

kenning: Confirming your thoughts on the overall status of monorepo/cms ds integration

roblevintennis:
@kenning yeah, she's just going to be utilizing the `restrict` method and nothing more. Should be clear if you look at her PR: getethos/ethos#1139

PR #22, opened 2019-09-26
[Prototype form component] Form I

kenning:
**For the quickest view of what this adds, pull this, run styleguide, and check out its page for Form.**

### Additions + changes: 
- Added `components/Form` and its hook, `hooks/useFormState`
  - Takes in `config` object which pretty much has all of the form info in it
  - Tracks state and any errors, wraps onsubmit, but lets inputs control their own state
  - Passes down form validity so the submit button can be disabled or enabled
  - Submit failure shows a form-level error
- Added `components/Form.md` with a working `config` prop
  - Has two inputs, both TextInputs, with different validation schemes
  - Submitting has a 50% chance to error out, handles errors and successes visibly for user
- Changed `hooks/useMinMaxLength` to go in a new folder, 'validators,' and changed name to `validateMinMax` (because it's not a hook)
- Add simple `validateTruthy` validator
- Some changes to TextInput 
  - min/max length validation is no longer baked into the component, fixing its behavior of handling errors and blurring differently from a passed-in validator
  - updated `components/inputs/TextInput.md` to have validateMinMax passed in
  - possible callback for registering errors (and lack of errors) in form wraps "setError" function, making sure that the visual error within an input is always up to date with the actual error status of the component. inputs still work fine if this isn't supplied!
    - Clicking submit before blurring component doesn't "trick" the form into submitting if it has errors

### Other changes I want to make after this but people may disagree:
- Make every `validator` prop on an input `validators` and have it take in an array of validators
- change "inputs" folder name to "fields," change all instances of "input" in new Form component to "field," because we need a word for "anything which counts as part of a form" (basically a different word which refers to both 'input' elements and 'select' elements

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -78,10 +78,6 @@ export function Form({ children, config, render }) {
 
     try {
       await config.onSubmit(getInputValues())
-      alert(
-        'form submission successful with values:' +
-          JSON.stringify(getInputValues())
-      )
     } catch (e) {
       setFormErrorMessage('Something bad happened: ' + e.toString())
     }
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -1,0 +1,67 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { useFormState } from '../../hooks/useFormState'
+
+export function Form({ children, config, render }) {
+  const inputNames = Object.keys(config.inputs)
+
+  // Set up initial values
+  let initialValues = {}
+  inputNames.forEach((x) => {
+    initialValues[x] = null
+  })
+
+  // Hooks
+  const [
+    getInputErrors,
+    setErrorStateFactory,
+    getFormErrorMessage,
+    setFormErrorMessage,
+    getFormIsValid,
+  ] = useFormState(initialValues)
+
+  // Validates form, sets button to disabled if there's an error
+  function inputNamePropsTransformer(inputName) {
+    return {
+      name: inputName,
+      formErrorHandler: setErrorStateFactory(inputName),
+      validator: (input) =>
+        config.inputs[inputName].validators
+          .reduce((errors, validator) => errors.concat(validator(input)), [])
+          .filter((x) => !!x) // remove empty strings
+          .join(', '),
+      labelCopy: config.inputs[inputName].labelCopy,
+      ['data-tid']: [config.formName, config.formId, inputName].join('-'),
+    }
+  }
+
+  async function onSubmit(syntheticReactEvent) {
+    console.log('syntheticReactEvent', syntheticReactEvent)
+    syntheticReactEvent.preventDefault()
+
+    if (!getFormIsValid()) {
+      if (getInputErrors()) {
+        setFormErrorMessage(
+          'One of the inputs was invalid. Errors: ' + getInputErrors()
+        )
+      } else {
+        setFormErrorMessage("You haven't typed anything yet")
+      }
+      return
+    }
+
+    try {
+      await config.onSubmit(syntheticReactEvent)
+      alert('form submission successful')
+    } catch {
+      setFormErrorMessage('Form was invalid')
+    }
+  }
+
+  return (
+    <form onSubmit={onSubmit}>
+      {children(inputNamePropsTransformer, getInputErrors, getFormErrorMessage)}
+    </form>
+  )
+}
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -3,6 +3,28 @@ import PropTypes from 'prop-types'
 
 import { useFormState } from '../../hooks/useFormState'
 
+/**
+ * The `Form` component takes a `config` object, and uses the convention
+ * of providing a callback function that you can place your `children`
+ * JSX in. See corresponding markdown example for details.
+ *
+ * The callback provides the following _parameters_:
+ * `inputNamePropsTransformer`—takes `inputName` and:
+ * - Validates form
+ * - Sets button to disabled if there's an error
+ *
+ * `getInputErrors`—TBD
+ *
+ * `getFormErrorMessage`—Provides for "form level errors" e.g. for
+ * presenting API errors, or messaging that relates to the entire form.
+ * _Note that field level errors (or hints if you will) are handled via
+ * the `validators` at the field level._
+ *
+ * `getFormIsValid`—Call this to determine if the form is generally valid.
+ * Useful for determining whether to disable a submit button for example.
+ *
+ * @see See also: `src/hooks/useFormState.js`
+ */
 export function Form({ children, config, render }) {
   const inputNames = Object.keys(config.inputs)
 
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -9,7 +9,7 @@ export function Form({ children, config, render }) {
   // Set up initial values
   let initialValues = {}
   inputNames.forEach((x) => {
-    initialValues[x] = null
+    initialValues[x] = 'err' // by default inputs have "hidden" errors
   })
 
   // Hooks
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -15,7 +15,8 @@ export function Form({ children, config, render }) {
   // Hooks
   const [
     getInputErrors,
-    setErrorStateFactory,
+    getInputValues,
+    setStateFactory,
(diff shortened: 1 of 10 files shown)


src/components/Inputs/TextInput/TextInput.js
```diff
@@ -54,26 +48,28 @@ function PrivateTextInput({
...
 
   const [value, setValue] = useState('')
 
-  const doValidation = (value) => {
-    const minMaxError = minMaxValidator(value)
-    if (minMaxError) {
-      setError(minMaxError)
-      return
+  const setErrorWrapper = (errorValue) => {
+    if (!!formChangeHandler) {
```
roblevintennis: Is formChangeHandler leftover? I don't see in proptypes or example. If not, can you add those please?

src/components/Form/Form.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  async function onSubmit(syntheticReactEvent) {
+    setFormErrorMessage('')
+
+    syntheticReactEvent.preventDefault()
+
+    if (!getFormIsValid()) {
+      // Should already be handled by form itself, but...
+      if (getInputErrors()) {
```
roblevintennis: Can you add an example of `getInputErrors` or put some 
```js
//example code I can plop into styleguidist live editor here??
```

src/hooks/useFormState.js
```diff
@@ -0,0 +1,71 @@
...
+    return touched && !getInputErrors()
+  }
+
+  return [
+    getInputErrors,
+    getInputValues,
+    setStateFactory,
+    getFormErrorMessage,
+    setFormErrorMessage,
+    getFormIsValid,
```
roblevintennis: So this is really `getFormIsValidAndTouched` in terms of what it really means. Your names are already seem to prefer being longer and expressive which is cool, but `getFormIsValidAndTouched` is getting into objective-c length haha. Do you have any ideas on something a bit more informative for this particular one?

Also, I was thinking we could lose the `Form` parts of these, but I realize it helps a lot to distinguish between fields within the jsx so I think you're right to have that 👍

src/components/Form/Form.md
```diff
@@ -0,0 +1,66 @@
...
+
+      <Spacer.H16 />
+
+      {getFormErrorMessage() && (
+        <InfoMessage.Alert.Error>
+          {getFormErrorMessage()}
+        </InfoMessage.Alert.Error>
+      )}
+
+      <Button.Medium.Black disabled={!getFormIsValid()} type="submit">
```
roblevintennis: yeah, this seems pretty nice and straight forward 👍

src/components/Form/Form.md
```diff
@@ -0,0 +1,66 @@
+```jsx
+import { TextInput, Spacer, Button, InfoMessage } from '../index'
+import validateMinMaxFactory from '../../validators/validateMinMax'
+import validateTruthy from '../../validators/validateTruthy'
+;<Form
+  config={{
+    formName: 'Styleguidist example form',
+    formId: '1',
+    inputs: {
```
roblevintennis: ok, I can sort of see why you want to move towards `^input^field`. Do we then consider certain, ahem, fields NOT fields e.g. I suppose here a `<Button />` is not considered a "field" per se. I suppose we'd want to be crystal clear what a field is so we can make it easy for team to understand. I suppose you could say "If you have something like a button that has no `validator` or `TBD???`, we consider that a component and not really a field." Let's contemplate how to make this clear.

roblevintennis:
Nice job @kenning I pretty much like this overall esp. since it was meant to be a POC and I feel like it's pretty understandable and loosely coupled. Seems to mainly meet the constraints we discussed 👍 

Next steps (I know you only had a few hours! But, just capturing these thoughts):

1. We need to expose the `useForm` in the docs. I think it might be easiest to comment those in at the component level (even if it means some duplication). react-styleguidist docs are pretty clear on how to get those props to show up as a table.
2. I left some random comments and thoughts we can address. I don't see any of these as huge concerns, so I think we'd be able to iterate on those (also haven't yet looked at form-II)
3. I think I should maybe add some test coverage for these and it'll help me understand better and find anything that needs refining or what not
4. I think given the importance of this abstraction I'd like @zach-ethos to have a look too

src/components/Form/Form.js
```diff
@@ -0,0 +1,78 @@
...
+      )
+    } catch (e) {
+      setFormErrorMessage('Something bad happened: ' + e.toString())
+    }
+  }
+
+  return (
+    <form onSubmit={onSubmit}>
+      {children(
+        inputNamePropsTransformer,
```
roblevintennis: This is where I'm thinking we can hook into react-styleguidist and gets some nice docs explanation going here. I think it'll take java comment arbitrarily (maybe?). Alternatively, line 17 could have these instead. To restate, I think understanding these callback params will be really important for users.

src/components/Inputs/TextInput/TextInput.js
```diff
@@ -54,26 +48,28 @@ function PrivateTextInput({
 
   // Set up validation hooks
   const [getError, setError, validate] = useErrorMessage(validator)
-  const [minMaxValidator] = useMinMaxLength(minLength, maxLength)
```
roblevintennis: yah, I like how we've consolodated all these into the validator(s) 👍

src/components/Inputs/TextInput/TextInput.md
```diff
@@ -1,13 +1,16 @@
...
   name="example"
-  minLength={5}
-  maxLength={20}
-  allCaps={true}
   labelCopy="Validation happens after first blur ('touched')—Value's length % 2"
   data-tid="the-text-input"
-  validator={(x) =>
-    x.length % 2 ? 'Text does not have an even number of characters' : ''
-  }
+  validator={(x) => {
```
roblevintennis: haha, so our 6:00-8:00pm conversation WAS helpful 😋

src/components/Inputs/TextInput/TextInput.js
kenning: didn't do proptypes for these yet because this is a prototype. formChangeHandler gets injected here: https://github.com/getethos/ethos-design-system/pull/22/files#diff-c17788385171ff0dfafebf350fa8aa26R51

if that makes sense, please resolve this conversation

src/components/Form/Form.js
kenning: sure one sec

kenning: well i'll do it in form-ii

src/components/Form/Form.md
kenning: I think the folder structure is enough. I moved button select group to the "inputs" folder, with your blessing i'd like to just change that folder to be called "fields" and that should be enough. Tell me if you want this change to get made. 

`<Button.Medium.Black type="submit">Submit</Button.Medium.Black>` feels very nice to me. it's a form so submit should work right?

src/components/Form/Form.js
kenning: did it over there, basically the submit isn't set to display as `disabled={true}`. In this example form its disabled as long as the form is untouched and has errors.

src/components/Form/Form.js
kenning: like we said on a call i think i'll just modify form-ii and we can work off that

src/components/Inputs/TextInput/TextInput.js
kenning: thx yea that was bugging me

src/hooks/useFormState.js
kenning: the form shouldn't be valid if it's not touched... idk man i think this is ok. we can change it later if we come up with something better?

src/components/Inputs/TextInput/TextInput.md
kenning: this is pretty much the main case for `validator` turning into `validators` and accepting an array. checking for truthiness is basically step 1 for marking a text input as required, but obviously it wouldn't work on a input which accepts numbers from 0-x. maybe it's a little verbose and requires excessive work from the client (i dont think so) but i sorta like the "composing a list of validator functions" thing and it will probably be appealing to the FP fans out there.

src/components/Inputs/TextInput/TextInput.js
kenning: it gets inserted here: https://github.com/getethos/ethos-design-system/pull/22/files#diff-c17788385171ff0dfafebf350fa8aa26R51

PR #23, opened 2019-09-26
[Prototype form component] Form II (jsonable!)

kenning:
**For the quickest view of what this changed, check out the styleguide page for Form. The difference between this and https://github.com/getethos/ethos-design-system/pull/22 is pretty clear.**

Same as https://github.com/getethos/ethos-design-system/pull/22 except now everything except the onSubmit of the form is defined with 100% json-compatible syntax. It's pretty readable too

### DIFF
--- a/src/components/Form/ComponentMap.js
+++ b/src/components/Form/ComponentMap.js
@@ -1,0 +1,11 @@
+import React from 'react'
+import { TextInput } from '../index'
+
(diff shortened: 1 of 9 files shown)


kenning:
Tried to delete this and remake as a PR on top of form-I but they diverged too much. Go to #24 instead

PR #25, opened 2019-09-26
[Prototype form component] Form II -- jsonable!

kenning:
Same as https://github.com/getethos/ethos-design-system/pull/22 except now everything except the onSubmit of the form is defined with 100% json-compatible syntax. It's pretty readable too


src/components/Form/Form.js
```diff
@@ -23,20 +25,6 @@ export function Form({ children, config, render }) {
   ] = useFormState(initialValues)
 
   // Validates form, sets button to disabled if there's an error
-  function inputNamePropsTransformer(inputName) {
```
kenning: This basically turns into "input" function below

src/components/Form/ValidatorMap.js
```diff
@@ -0,0 +1,16 @@
+import validateMinMaxFactory from '../../validators/validateMinMax'
+import validateTruthy from '../../validators/validateTruthy'
+
+export default function GenerateValidator(validatorName, args) {
+  switch (validatorName) {
+    case 'truthy':
+      return validateTruthy
+    case 'minMax':
+      return validateMinMaxFactory.apply(null, args)
+    case 'exampleEvenNumber':
```
kenning: Throwaway code for the purposes of illustrating what this setup can do.

src/components/Form/ComponentMap.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import { TextInput } from '../index'
+
+export default function GenerateComponent(componentName, props) {
```
roblevintennis: Ok, I'd call this a `ComponentGenerator` fwiw, but, I think for this sort of factory thing to be useful you MUST keep a library of widgets that matches your components. So it violates open/closed principle (google it) in that you'll have to keep it up to date by adding/removing components from your switch state here (I presume this is incomplete for POC purposes. But, it _would_ need to potentially support all components.

roblevintennis: Think about how shitty kitchensink started being to keep up to date for example-- that's the sort of high maintenance I see this becoming

src/components/Form/ComponentMap.js
```diff
@@ -0,0 +1,11 @@
...
+  switch (componentName) {
+    case 'TextInput':
+      return <TextInput {...props} />
+    default:
+      throw new Error('whoopsie daisy')
```
roblevintennis: So this confirms to my mind that your intent is you can only ask for components that exist. So my assertion above about having to maintain the list seems accurate.

src/components/Form/Form.js
```diff
@@ -58,21 +46,33 @@ export function Form({ children, config, render }) {
...
+          "\n\nBut try again, it's random"
       )
     } catch (e) {
       setFormErrorMessage('Something bad happened: ' + e.toString())
     }
   }
 
+  function input(inputName) {
+    const inputConfig = config.inputs[inputName]
+    return ComponentMap(inputConfig.componentName, {
```
roblevintennis: Ok, I see how your using this. So...

`ComponentMap` to my mind should be controlled by the client. This way, the client can assemble just the "UI widgets" it needs and circumnavigates my concerns about maintaining a big switch case list of components. So, I have a form and I want a `TextInput` and `ButtonSelectGroup`...in that case, I can assemble my own component map on `config.map` (and even validators with `config.validators`).

I think you could do ^ without really changing the spirit of what you're doing here, no?

If I had any mandate, I'd say inverting the control on the maps and validators is my biggest concern.

src/components/Form/Form.js
```diff
@@ -58,21 +46,33 @@ export function Form({ children, config, render }) {
...
+          JSON.stringify(getInputValues()) +
+          "\n\nBut try again, it's random"
       )
     } catch (e) {
       setFormErrorMessage('Something bad happened: ' + e.toString())
     }
   }
 
+  function input(inputName) {
+    const inputConfig = config.inputs[inputName]
```
roblevintennis: might as well start using the vernacular fields for all of this no? Or is that v2b? :)

roblevintennis: But that's the direction you have in mind for later yah??

src/components/Form/ValidatorMap.js
roblevintennis: Yeah, I'd say my comments for the "mappers" and "validators" can be considered the same. I don't want us to have to manage/maintain all these. I think it's enough for us to provide some boiler-plate validators, but this generation/builder type of stuff belongs higher up layer and on the client side to my mind.

kenning: sounds good

src/components/Form/ComponentMap.js
kenning: yep will push changes

PR #28, opened 2019-09-26
Form Component

kenning:
**For the quickest view of what this adds, pull this, run styleguide, and check out its page for Form.**

### Additions + changes: 
- Added `components/Form` and its hook, `hooks/useFormState`
  - Takes in `config` object which pretty much has all of the form info in it
  - Tracks state and any errors, wraps onsubmit, but lets inputs control their own state
  - Passes down form validity so the submit button can be disabled or enabled
  - Submit failure shows a form-level error
- Added `components/Form.md` with a working `config` prop
  - Has two inputs, both TextInputs, with different validation schemes
  - Submitting has a 50% chance to error out, handles errors and successes visibly for user
- Changed `hooks/useMinMaxLength` to go in a new folder, 'validators,' and changed name to `validateMinMax` (because it's not a hook)
- Add simple `validateTruthy` validator
- Some changes to TextInput 
  - min/max length validation is no longer baked into the component, fixing its behavior of handling errors and blurring differently from a passed-in validator
  - updated `components/inputs/TextInput.md` to have validateMinMax passed in
  - possible callback for registering errors (and lack of errors) in form wraps "setError" function, making sure that the visual error within an input is always up to date with the actual error status of the component. inputs still work fine if this isn't supplied!
    - Clicking submit before blurring component doesn't "trick" the form into submitting if it has errors

### Other changes I want to make after this but people may disagree:
- Make every `validator` prop on an input `validators` and have it take in an array of validators
- change "inputs" folder name to "fields," change all instances of "input" in new Form component to "field," because we need a word for "anything which counts as part of a form" (basically a different word which refers to both 'input' elements and 'select' elements


roblevintennis:
LGTM

image


image

PR #30, opened 2019-09-27
Birthdate field works in `<Form>`

kenning:
Check form.md for an example

**Working form**
x
**Working form with error handling**
x
**Working original birthdate input**
x

### DIFF
--- a/src/validators/BirthdateInputValidator.js
+++ b/src/validators/BirthdateInputValidator.js
@@ -1,11 +1,6 @@
-import React from 'react'
-import dayjs from '../../../helpers/getDayjs.js'
+import dayjs from '../helpers/getDayjs.js'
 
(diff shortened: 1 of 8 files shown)


src/validators/BirthdateInputValidator.js
```diff
@@ -50,34 +41,26 @@ export const getMaxDateValidator = (props) => {
...
 }
 
-export const getMinMaxDateValidator = (props) => {
-  const {
-    minAge,
-    maxAge,
-    dateFormat,
-    customErrorMessage,
-  } = props;
+export const validateMinMaxDateFactory = (props) => {
```
kenning: Massive change here from prettier, the only thing i actually changed in this file was the name of this exported validator.

PR #32, opened 2019-09-27
Add system serif & update sans-serif font stacks for font display fallback/swap

c7hudson:
See request for updated stacks from Design team here: https://getethos.slack.com/archives/CMMQV98JC/p1568667139001900
```
Sans Serif:
1. Theinhardt
2. SF (San Francisco / -apple-system, BlinkMacSystemFont)
3. Helvetica Neue
4. Helvetica
5. Arial
6. Sans Serif

Serif:
1. Cambon
2. Georgia
3. Times New Roman
4. Serif
```

Initially implemented on CMS side before we were using DS as a package. Required for expected performance of `font-display` in CMS.

Should note that San Francisco will only be used in cases where it's already installed on the user's system, since we're not trying to download another web font here for users who don't have it.


roblevintennis:
@c7hudson @zach-ethos to save me some code diving time can you explain to me when we fallback on `--system-` stacks exactly?

c7hudson:
@roblevintennis if the browser fails to load our brand fonts quickly or at all it will fall back to the next in the stack.

With our serif text we'll eventually always swap to our brand font once it's loaded, but will show a fallback if not immediately downloaded, so the text is still readable. 

For our sans serif, if it doesn't download within a certain period of time it'll use the fallback instead, so the user doesn't have a jarring experience of the font changing while they're reading longer body copy (as opposed to a shorter serif heading).

We discussed our implementation of `font-display` `fallback` and `swap` with design and they chose these as the fallbacks.

We didn't have a serif fallback stack so that was the main reason for implementing this but it also revised the original sans-serif stack. That way if there is a font swap event it's less noticeable than jumping serif to sans serif.

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

Related pr where we add font-display on CMS https://github.com/getethos/cms/pull/1729

c7hudson:
Sorry I should have given this initial context with more detail when opening the PR.

roblevintennis:
I just approved. thanks for context @c7hudson ...  mind pointing me to the code that mechanically makes the switching to system fallback after certain allotted time as a github link please? Just would like to have a look for my own understanding. Thanks sir!

c7hudson:
@roblevintennis I've added the built files for this PR - didn't catch that I forgot to commit when I opened.

Cambon stack `font-display` usage: https://github.com/getethos/cms/blob/master/src/components/styles/style.scss#L19

Theinhardt stack `font-display` usage:
https://github.com/getethos/cms/blob/master/src/fonts/Theinhardt-Light.css#L17
https://github.com/getethos/cms/blob/master/src/fonts/Theinhardt-Medium.css#L17
https://github.com/getethos/cms/blob/master/src/fonts/Theinhardt-Regular.css#L17

The allotted times are controlled depending on browser as far as I know, this is an older article but it has a table of browser support that includes the timeout periods: https://developers.google.com/web/updates/2016/02/font-display#differences_in_font_rendering_today

This illustration shows a comparison of different `font-display` options in regards to their timelines.
be746fa6-6ce9-423d-b9a3-684949adc8d5_Screen Shot 2017-11-14 at 2 49 37 PM

ref: https://font-display.glitch.me/

PR #34, opened 2019-09-30
Qw II tid and no "get()" for validators

kenning:
Screenshot of `tid` getting passed down from form config to an input:
x


kenning:
CMS branch: https://github.com/getethos/cms/pull/1846

PR #40, opened 2019-10-02
New <Image>, responsive, lazy-loading, retina-aware, automatically uses best image format

kenning:
**ToDo**
- [ ] test with CMS
- [ ] setup multiple source values inside picture tag (maybe)
- [ ] fetch_format="auto" (if extension != svg)
- [ ] quality="auto:eco"
- [ ] We need to include lazysizes.min.js in the webpack bundle for this to actually work.

**Description:**
- [Asana Task](https://app.asana.com/0/1116481661798430/1142171941519789/f)

Need help from cole on this one. I think i basically accomplished everything in the asana task, and I think this should work with the CMS as it is right now... 

Not sure how exactly to test lazy loading, but it should work properly (lazysizes.min.js is included in styleguidist, and the images have the 'lazyload' class)


src/components/Images/Images.js
```diff
@@ -1,31 +1,81 @@
...
+      fetchFormat: 'auto',
+      srcset: {
+        breakpoints: [600, 900, 1200],
+        sizes: true,
+      },
+      dpr: 'auto',
+      crop: 'fill', // ?
+      secure: 'true', // ?
+    },
+    !!width ? { width } : {},
```
anbright: Can we use `Boolean()` here?

src/components/Images/Images.js
```diff
@@ -1,31 +1,81 @@
...
 // https://cloudinary.com/documentation/image_transformation_reference
 
 export const CLOUDINARY_CLOUD_NAME = 'getethos'
 
-const publicIdBase = `https://res.cloudinary.com/${CLOUDINARY_CLOUD_NAME}/image/upload/`
+const cld = new cloudinary.Cloudinary({
+  cloud_name: CLOUDINARY_CLOUD_NAME,
+})
+
+const BREAKPOINTS = [600, 900, 1200]
```
anbright: Do these correspond to the Media queries? If not, should we store them in a `constants.js` file somewhere?

src/components/Images/Images.js
kenning: now i'm curious, looking up which is faster. This will get called twice for every image so we should do whatever's faster

kenning: Looks like `!!` is faster. http://jsben.ch/JWVSk http://jsben.ch/nko0g The difference is pretty small, so if you prefer `Boolean()` we can change it.

src/components/Images/Images.js
kenning: They correspond to media queries.

kenning: I dont actually use them anymore 😅good catch

kenning: I don't actually use them, instead I redefine the same thing later 😅 good catch

src/components/Images/Images.js
```diff
@@ -1,64 +1,114 @@
...
+  )
+  includesInvalid(rest)
+
+  const srcSetOptions = Object.assign(
+    {},
+    {
+      // Find information on image transformations here:
+      // https://cloudinary.com/documentation/image_transformations
+      quality: 'auto:eco',
+      fetchFormat: 'auto',
```
kenning: @c7hudson this look ok?

src/components/Images/Images.js
roblevintennis: Drew I’m seeing this without access to my laptop ATM...is this the direction we want to go?

kenning:
@roblevintennis

src/components/Images/Images.js
roblevintennis: Cole any update on the direction we want to go per last week’s investigation? I’m not on my laptop ATM

roblevintennis: Cole any update on the direction we want to go per last week’s investigation? I’m not on my laptop ATM

c7hudson: @roblevintennis Yes, here is what I've brainstormed: https://www.notion.so/getethos/CloudinaryImage-Refactor-Brainstorming-03aa0973e2674d689779c84b279eba7f

It'll take most of what Nick has done here into consideration and I'll build from it. There have been some issues with fetchFormat not working properly, which is why I'm suggesting breaking things out into explicit requests for image formats.

c7hudson: Although if I can get fetch format and dpr auto working consistently, then the brainstorming method I've posted above can be simplified a bit.

roblevintennis: @c7hudson are you planning to literally cherry-pick in commits from this or just use it for reference?

@kenning was asking me what the status was on this one, so I think from above input I'll let you guys work this out initially.

Questions: Is `<CloudinaryImage />` and `<picture />` two separate approaches we're considering? Or, is the former wrapping the later?

Also, can you give me a short "helicopter overview" of what Cloudinary image transformations are? I sort of got lost in their docs. I presume it's CDN + automagical image optimization of some kind, right? I just don't yet have any experience with how all this works. Thanks!

c7hudson: @roblevintennis I plan to PR into this PR with some additional work, as well as setup a PR CMS side to change how we implement `CloudinaryImage`. I could setup a new branch to base off of the latest `master` version of EDS as well.

`<CloudinaryImage/>` would render the `<picture/>` element.

The way we use image transformations right now, we create a URL for the source of an image that tells Cloudinary what we want delivered in terms of image dimensions, format, quality, device pixel ratio and cropping technique. The transformations we request are delivered via the Cloudinary CDN and cached on their end so subsequent requests for the same transformed image do not have to initiate a new transformation.

Right now Cloudinary runs a bit of JS, through the `cloudinary-react` package that we rely on, to determine the correct DPR and when using fetch format it's supposed to serve next gen image formats (although I haven't seen it serve `.jp2` yet and it caused issues with our SVGs). 

We also wrap most `CloudinaryImage` elements in `Media` components in the CMS, something we're trying to get away from due to the instability of `<Media>`. 

So moving to `<picture>` removes the need for Cloudinary to run any client detection scripts for determining next gen image format or DPR, and through use of the `media` attribute in the `<source>` tags we can remove the need for the `<Media>` wrappers. 

Overall this should reduce the amount of JS involved in serving images and also rely a bit less on the "automatic magic" of Cloudinary, by being more explicit with the URLs we request for our images.

c7hudson: We also weren't able to impement lazy loading using `cloudinary-react` because it doesn't provide a way to output an `img` tag with anything other than a `src` attribute for the image URL. So you can't do `data-src` which most lazy loading techniques rely on. Moving to the core Cloudinary JS SDK allows us to generate resource URLs and use them in whatever markup we desire.

roblevintennis: Really helpful overview, thanks sir 🙏

c7hudson:
closing - see #53

PR #42, opened 2019-10-03
Qw ii rz cleanup form interacted with

roblevintennis:
# For now pointed to qwII-rz-cleanup but after that's merged we'll point this to master

___

**Description:**
- [Asana Task](https://app.asana.com/0/1137984670709817/1142882503761970/f)
- Form interacted with is now exposed to consumer

- 

**Screenshots:**
image

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -61,6 +61,7 @@ export function Form({ children, config }) {
     getFormErrorMessage,
     setFormErrorMessage,
     getFormIsValid,
+    getFormInteractedWith,
   ] = useFormState(initialValues)
 
   // Wrapper for default <form> submit function.
@@ -138,7 +139,12 @@ export function Form({ children, config }) {
     <form onSubmit={onSubmit}>
       {/* See the top of this file or ./Form.md for help on how to use
         these arguments passed to the children function. */}
-      {children(field, getFormErrorMessage, getFormIsValid)}
+      {children(
+        field,
+        getFormErrorMessage,
+        getFormIsValid,
+        getFormInteractedWith
+      )}
     </form>
   )
 }
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -117,7 +117,7 @@ export function Form({ children, config }) {
           fieldConfig.validators
             .reduce((errors, validator) => errors.concat(validator(field)), [])
             .filter((x) => !!x) // remove empty strings
-            .join(', '),
+            .join('. '),
 
         // User-visible copy, shows up in the label above the field.
         labelCopy: fieldConfig.labelCopy,
--- a/src/components/Form/Form.md
+++ b/src/components/Form/Form.md
@@ -1,9 +1,8 @@
-_Note that we've set up the form submission to randomly fail or succeed—so, you're encouraged to resubmit until you've seen both!_
-
 ```jsx
 import validateTruthy from '../../validators/validateTruthy'
 import validateMinMaxFactory from '../../validators/validateMinMax'
 import { TitleLarge, TextInput, Spacer, Button, InfoMessage, ZipInput, ZipInputValidator } from '../index'
+let count = 0
 ;<Form
   config={{
     formName: 'Styleguidist example form',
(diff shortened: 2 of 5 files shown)


src/components/Inputs/ButtonSelectGroup/ButtonSelectGroup.md
```diff
@@ -33,7 +33,7 @@
 ```jsx
 <ButtonSelectGroup
   defaultValue="excellent"
-  label="Health"
+  labelCopy="Health"
```
roblevintennis: This was that button select group error we kept seeing

src/hooks/useFormState.js
```diff
@@ -29,6 +29,10 @@ export function useFormState(initialState) {
     }
   }
 
+  function getFormInteractedWith() {
+    return touched
```
roblevintennis: I double checked `setStateFactory` which is where this is set and basically the first time a button select group is clicked, or, a character is entered into a field (text, birthdate, zip) `touched` (for the form) becomes true. So, this works for our purposes.

I was thinking of exposing `getFieldValues` but Nick pointed out that was meant to be a private variable, and also I was thinking if we somehow checked all field values are `null` to determine if the form was interacted with, that could get pretty inefficient with a form that has a lot of fields.

Happy that we already had a flag to check and expose this.

src/components/Form/Form.js
```diff
@@ -116,7 +117,7 @@ export function Form({ children, config }) {
           fieldConfig.validators
             .reduce((errors, validator) => errors.concat(validator(field)), [])
             .filter((x) => !!x) // remove empty strings
-            .join(', '),
+            .join('. '),
```
roblevintennis: I've  removed the double error messaging  we saw on Zip, but, in the event we do have multiple, these should be sentances e.g. "Please do this. Please do that."

src/components/Form/Form.js
```diff
@@ -138,7 +139,12 @@ export function Form({ children, config }) {
     <form onSubmit={onSubmit}>
       {/* See the top of this file or ./Form.md for help on how to use
         these arguments passed to the children function. */}
-      {children(field, getFormErrorMessage, getFormIsValid)}
+      {children(
+        field,
+        getFormErrorMessage,
+        getFormIsValid,
+        getFormInteractedWith
```
roblevintennis: @iamvictorli this will unblock you

src/components/Form/Form.md
```diff
@@ -39,24 +38,31 @@ import { TitleLarge, TextInput, Spacer, Button, InfoMessage, ZipInput, ZipInputV
     },
     onSubmit: async (formData) => {
       await new Promise((resolve) => setTimeout(resolve, 500))
-      if (!!(Math.floor(Math.random() * 10) % 2)) {
-        throw new Error("Oh no, the api is broken (try again, it's random)")
+      if (count++ % 2 === 0) {
```
roblevintennis: Multiple requests to make this  deterministic (sorry @kenning  😛 cc / @anbright @ekeric13 I believe y'all provided this feedback)

src/components/Form/Form.md
```diff
@@ -39,24 +38,31 @@ import { TitleLarge, TextInput, Spacer, Button, InfoMessage, ZipInput, ZipInputV
...
   }}
 >
-  {(field, getFormErrorMessage, getFormIsValid) => (
+  {(field, getFormErrorMessage, getFormIsValid, getFormInteractedWith) => (
     <div>
       <TitleLarge.Serif.Book500>Example Form</TitleLarge.Serif.Book500>
 
       <Spacer.H16 />
 
+      {!!getFormInteractedWith() && (
```
roblevintennis: @iamvictorli just an example usage, but you will easily be able to call:
```js
      {!!getFormInteractedWith() && ( removeCTAButton() }
```

src/components/Form/Form.js
iamvictorli: 👍

src/components/Form/Form.md
roblevintennis: and you probably don't need the `!!` part

iamvictorli: @roblevintennis You probably dont need the `!!` operator, since the getFormInteractedWith already returns a boolean

roblevintennis:
I'm going  to merge this as it's pointed to another  topic branch anyway. Also, I won't to have it's base branch mergable soon, and so I'll need to point CMS to the latest here.

PR #45, opened 2019-10-04
Buttons Group Select not properly using validators

roblevintennis:
**Description:**
- [Asana -- Buttons Group Select not properly using validators](https://app.asana.com/0/1137984670709817/1143200052992145/f)
- This also hopefully makes the comments around all this clearer so we can better understand what's going on. 
- Birthdate does prevalidation and only kicks in `validators` once we have a valid date format string. This tries to call that out and make it more immediately understandable


roblevintennis:
cc / @anbright @zach-ethos I'm going to merge this and try to work on CMS side on my train ride in. The button select group definitely wasn't "wired up" and using our validate pattern. This should help with that and also make the markdown demo much more helpful.

src/components/Form/Form.js
```diff
@@ -97,6 +101,15 @@ export function Form({ children, config }) {
     // This just makes the rest of this easier to read
     const fieldConfig = config.fields[fieldName]
 
+    const doValidation = (field) => {
```
roblevintennis: Perhaps more verbose, but just moving this out made this much easier to debug so I opted to leave it like this. Functionality is the same as what Nick original had.

src/components/Form/Form.md
```diff
@@ -3,6 +3,12 @@ import validateTruthy from '../../validators/validateTruthy'
 import validateMinMaxFactory from '../../validators/validateMinMax'
 import { TitleLarge, TextInput, Spacer, Button, InfoMessage, ZipInput, ZipInputValidator } from '../index'
 let count = 0
+
+function validateCustom(x) {
```
roblevintennis: This will allow consumers / developers to see more clearly how these kick in or not

src/components/Form/Form.js
```diff
@@ -97,6 +101,15 @@ export function Form({ children, config }) {
...
+      return fieldConfig.validators
```
roblevintennis: It's crazy to me that this same exact code is being done in the `analyticsWrapper` and we're not just passing in the analytics event we want called and reconciling here :/

PR #46, opened 2019-10-04
More permissive blacklist

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1137984670709817/1143293029720137/f)
- Restrict is too prohibitive

This came from call center folks admin via @mwwu


mwwu:
Rob merged before I approved, but approved!

PR #48, opened 2019-10-07
Provide Links As Props in UniversalNavBar

iamvictorli:
**Description:**
- Currently in the CMS, the navbar is not using `gatsby links` and using anchor tags
- `/term` link should not be used with gatsby-links as they are not an [internal link](https://www.gatsbyjs.org/docs/gatsby-link/#reminder-use-link-only-for-internal-links) for CMS.
- How this would be used`<UniversalNavbar LinkComponent={gatsby-link} />`

- [PR](https://github.com/getethos/cms/pull/1743)

**Screenshots:**


src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -108,16 +117,16 @@ class UniversalNavbar extends React.Component {
...
+                  <Link href="/">
                     {LogoWhite({ className: 'universal-navbar-logo' })}
-                  </a>
+                  </Link>
                   <Spacer.H56 />
                   {navbarLinks.map((l) => (
                     <div key={l.title + 'mobile'} style={{ marginBottom: 24 }}>
                       <TitleXLarge.Sans.Regular400>
-                        <a key={l.title + 'mobile'} href={l.href}>
+                        <Link key={l.title + 'mobile'} href={l.href}>
```
roblevintennis: What if two links have same title? I think this would break. See `src/components/Inputs/ButtonSelectGroup/OptionButton.js` which I believe is using uuid for aria labels, but, since we already have that dep you can use for keys too. I did this somewhere in CMS recently if you're looking for an example

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -50,6 +50,15 @@ class UniversalNavbar extends React.Component {
   }
 
   render() {
+    const Link = ({ href, ...props }) => {
```
roblevintennis: What happens if I wanna use Link somewhere else? I can't. Is there any reason to not make Link a sharable component of it's own?

iamvictorli: This `Link` is created from the props from UniversalNavbar. I guess we can create a Link HOC component/function.

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -50,6 +50,15 @@ class UniversalNavbar extends React.Component {
...
+      const PropsLink = this.props.link
+      if (PropsLink) {
+        return <PropsLink to={href} {...props} />
+      }
```
iamvictorli: created from props from UniversalNavbar

iamvictorli:
> Do we have accompanying tests for this change?

I have updated the snapshot testing for UniversalNavBar.

src/components/UniversalNavbar/UniversalNavbar.js
zach-ethos: I think is basically a good approach (wrap the injected component and route the props as needed), though I think it's nonperformance (at runtime) to create a new React component inside a render method, because each render the function will be recreated.

zach-ethos: nonperformant*

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -161,6 +169,7 @@ class UniversalNavbar extends React.Component {
 UniversalNavbar.propTypes = {
   hideMobileCta: PropTypes.bool.isRequired,
   trackCtaClick: PropTypes.func.isRequired,
+  link: PropTypes.node,
```
zach-ethos: Can we call this something more descriptive? I would suggest `LinkComponent`, and capitalize it to suggest it's a React component. You can then just call `<LinkComponent>` directly without the extra step above (PropsLink = link).

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -183,8 +183,11 @@ class UniversalNavbar extends React.Component {
 }
 
 UniversalNavbar.propTypes = {
+  /** Hide cta on mobile viewport */
   hideMobileCta: PropTypes.bool.isRequired,
+  /** Run analytics function when CTA Button gets clicked */
   trackCtaClick: PropTypes.func.isRequired,
+  /** agnotistic Reach and React Router Link */
```
zach-ethos: Thanks for the docs :)

PR #50, opened 2019-10-08
Fix floating hamburger bug by removing JS Media helpers

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1116481661798430/1140356378203525)
- The new UniversalNavbar uses the new, bad/deprecated Media helpers, which I think was the cause of the bug above. Yet another reason we should remove or limit these.
- This PR replaces those components with simple divs + equivalent Sass MQ helpers, which makes the diff look bigger than it is (it's mostly the phone and tablet wrappers changing).
- Added a snapshot test for posterity.

- https://github.com/getethos/cms/pull/1900
- https://deploy-preview-1900--ethos-cms.netlify.com

**Screenshots:**
- n/a


zach-ethos:
Also @roblevintennis @bwalker8038 we should talk about CSS modules again. Probably out of scope for this ticket, but I think we could make it work to just move back to module CSS per component, kill `design-system.css` and its imports, and then the CMS/etc. just (if we can make this work) would just import the React component (which would import its own stylesheet).

In the meantime, I just went with BEM-style names for the global classes here.

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -86,87 +83,74 @@ class UniversalNavbar extends React.Component {
...
         >
           <Layout.ScrollDetector>
-            <Media.PhoneOnly>
-              {/* Hamburger button */}
-              <div style={{ zIndex: 2, position: 'fixed', top: 24, right: 24 }}>
-                <TransformingBurgerButton
-                  showMobileMenu={showMobileMenu}
-                  clickHandler={this.toggleHamburger}
-                />
-              </div>
```
roblevintennis: I don’t love BEM fwiw but for now I think it’s fine and I’m glad we’re on same page re converting to css modules and losing design-system.css. I’d also like to consider removing any need for bundle and use ES6 imports for CMS but I may be missing some things

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -86,87 +83,74 @@ class UniversalNavbar extends React.Component {
...
-                      </Button.Medium.BlackOutline>
-                    </a>
+                <a href={LINKS.INDEX.href}>
+                  {LogoWhite({ className: 'universal-navbar-logo' })}
+                </a>
+                <Spacer.H56 />
+                {LINKS.NAVLINKS.map((l) => (
+                  <div key={l.title + 'mobile'} style={{ marginBottom: 24 }}>
+                    <TitleXLarge.Sans.Regular400>
+                      <a key={l.title + 'mobile'} href={l.href}>
```
roblevintennis: UUID instead maybe?

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -86,87 +83,74 @@ class UniversalNavbar extends React.Component {
...
+                        {l.title}
+                      </a>
+                    </TitleXLarge.Sans.Regular400>
                   </div>
+                ))}
+                <div style={{ position: 'absolute', bottom: 40 }}>
```
roblevintennis: Should we use a class instead of inline?

src/components/UniversalNavbar/UniversalNavbar.js
zach-ethos: > I don’t love BEM fwiw but for now I think it’s fine and I’m glad we’re on same page re converting to css modules and losing design-system.css. 

I see BEM as basically manual CSS modules, so like CSS Modules > BEM > no convention. In general I'm very pro, but I worried we would run into trouble trying to define CSS modules in one place and consume them elsewhere. But I think we can figure it out.

> I’d also like to consider removing any need for bundle and use ES6 imports for CMS but I may be missing some things

Agreed.

src/components/UniversalNavbar/UniversalNavbar.js
zach-ethos: Sure, though I didn't write this, or even modify it in this PR really. I see this PR as scoped to solving the bug, which is the hamburger button webkit glitch due to the Media helpers. All this diff is just indentation mostly.

zach-ethos: Also maybe @iamvictorli just did this? Saw a similar diff a minute ago.

src/components/UniversalNavbar/UniversalNavbar.js
zach-ethos: Yes, I try to avoid inline styles. Similar to the previous comment though, I didn't want to refactor this entire file for a hotfix.

zach-ethos: (I.e. these styles were already here, and I just moved the wrappers and indentation.)

zach-ethos: I did move a few inline styles to Sass when they were more directly related to this ticket, i.e. the wrappers elements.

src/components/UniversalNavbar/UniversalNavbar.js
roblevintennis: Hrm, yeah, I may have said the same thing on his PR haha. Let's call this non blocking. I think we may want to work on using uuid across places at some point but if there are no errors in console we have bigger things to fix as a priority. Just know that uuid is available also in CMS now if you'd like to use it.

PR #51, opened 2019-10-08
[QA] Remove webpack build step

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1141281370073110)
- While I was out, the original static method of getting DS code into the CMS seemed problematic, so we switched to using a public GitHub repo as the target of an package.json dependency. 
- This solved a couple problems and created a couple problems.
- One problem it created is the need to transpile ES6 DS source code modules into ES5. TLDR this PR removes the webpack build step from the DS, and has a corresponding CMS PR that sets up Gatsby to transpile just the source files it needs directly. This ends up being a lot more efficient.

- https://github.com/getethos/cms/pull/1904

**Screenshots:**

Pictured: the CMS JS bundle, with the DS webpack bundle highlighted in white on the left (934kb).

Capture d’écran 2019-10-07 à 18 11 17


package.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.4-n",
+  "version": "1.1.5-a",
   "description": "Ethos Design System v1",
-  "main": "./build/bundle.js",
+  "main": "./src/components/index.js",
```
zach-ethos: This has the effect that `import { Caption } from 'ethos-design-system'` should keep working with minimal changes elsewhere. When we want something else (e.g. a helper), I think the path still works as it does in your recent work in monorepo @roblevintennis.

package.json
```diff
@@ -57,10 +59,8 @@
     "react-text-mask": "^5.4.3",
     "sass-loader": "^8.0.0",
     "style-loader": "^1.0.0",
-    "text-mask-addons": "^3.8.0",
     "url-loader": "^2.1.0",
     "webpack": "^4.39.3",
```
zach-ethos: @roblevintennis Seems you still need this and the loaders for styleguidist.

package.json
roblevintennis: I see, and I'm guessing netlify uses `main` as it's entry point and that's why we'd see issues on netlify cms preview when we didn't rebundle and bump version (whereas locally CMS would pick up the changes).

Yeah, I like this whole direction 👍

PR #52, opened 2019-10-08
Css module 1

roblevintennis:
**Description:**
- CSS Modules for one component at a time (I think)
- All other css gets compiled as before

So here's example:

foo.module.css -- gets compiled with css-loader modules: true
foo.css -- gets compiled as before


zach-ethos:
> `.get-an-estimate .Button.Medium { /* DS styles overrides */ }`

Yeah, I think the ability of dev to target global selectors like that is ultimately a footgun and one of the better arguments for moving the DS into CSS modules (which is our norm elsewhere). cc @bwalker8038 for thoughts

src/components/Button/Button.small.shame.scss
```diff
@@ -0,0 +1,40 @@
+// ************************************************************** //
```
roblevintennis: This whole commit removes the one-off scss in: `/src/components/UniversalNavbar/UniversalNavbar.scss` and I think the commit message says it all:

> Adds a one-off Button.Small only intended for use in the UniversalNavbar. I've made it as clear as possible that it's not to be used elsewhere. Removes UniversalNavbar.scss specific scss code and instead used the component directly--needed so we can use CSS Modules across the entire Button components

This was approved for time being by Design as I verified all this with Tyler. Someday we may want a `Button.Small`, but for now, this let's us proceed.

webpack.config.js
```diff
@@ -31,22 +31,34 @@ module.exports = {
         loader: 'babel-loader',
       },
       {
-        test: /\.(s*)css$/,
+        test: /\.s?css$/,
         exclude: /node_modules/,
-        use: [
-          { loader: 'style-loader' },
+        // So foo.module.scss gets treated as a css module foo.css does not
```
roblevintennis: So it turns out that gatsby's webpack config checks for same thing I believe


image

and
E.g. in `cms/node_modules/gatsby/dist/utils/webpack-utils.js`:

image

src/components/Button/Button.module.scss
```diff
@@ -148,6 +154,24 @@ button,
   }
 }
 
+// Purposely very specific name WhiteCTA as this is currently
+// only used on the CMS hero for the CTA 'Check my price' button
+.Button.WhiteCTA {
+  background: var(--White);
+  padding: 16px var(--Space-32);
```
bwalker8038: Nit: We should have `--Space-16` var available

src/components/Button/Button.js
```diff
@@ -80,10 +83,14 @@ function PrivateButton({
...
 
+  const cssModulesClasses = classNames.map((klass) => {
+    return styles[klass]
+  })
+
   return (
     <button
       {...checked}
-      className={classNames.join(' ')}
+      className={cssModulesClasses.join(' ')}
```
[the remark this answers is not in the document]
roblevintennis: I dig it. I've made mines even shorter but thanks!

PR #55, opened 2019-10-08
Update validateMinMaxDateFactory to accept date minBirthdate and maxBirthdate args

jameshawkinsjr:
**Description:**
-----

- [Asana Task](https://app.asana.com/0/1135109139115239/1142356863120217/f)
- This makes `validateMinMaxDateFactory` more flexible by taking in arguments `minBirthdate` and `maxBirthdate`, since they might be calculated in different ways with different carriers.
- `minAge` and `maxAge` will still be included as arguments, so that you can display the ages you want in the error message.

-----

- The Quote Widget in the CMS will pass these new params as part of this PR: https://github.com/getethos/cms/pull/1921

**Screenshots:**
-----

Screen Shot 2019-10-08 at 4 47 45 PM
Screen Shot 2019-10-08 at 4 47 51 PM


src/validators/BirthdateInputValidator.js
```diff
@@ -56,20 +56,18 @@ export const getMinDateValidator = (props) => {
 }
 
 export const validateMinMaxDateFactory = (props) => {
-  const { minAge, maxAge, dateFormat, customErrorMessage } = props
+  const {
+    minBirthdate,
+    maxBirthdate,
+    minAge,
```
roblevintennis: So we only need these for the messaging huh? Bummer. But I can't think of another way.

jameshawkinsjr: I was thinking of calculating the age from the supplied `minBirthdate` and maxBirthdate` , but I thought it might just be cleaner to pass in the ages you want to be displayed.

PR #56, opened 2019-10-09
No LinkComponent for /login

iamvictorli:
**Description:**
- `/login` is not an internal link in the CMS
- uuid for keys and non link component for /login

**Screenshots:**

### DIFF
--- a/src/components/UniversalNavbar/UniversalNavbar.js
+++ b/src/components/UniversalNavbar/UniversalNavbar.js
@@ -126,10 +126,10 @@ class UniversalNavbar extends React.Component {
                       <NavLink
                         href={link.href}
                         LinkComponent={
-                          l.href !== '/login/' ? LinkComponent : null
+                          link.href !== '/login/' ? LinkComponent : null
                         }
                       >
-                        {l.title}
+                        {link.title}
                       </NavLink>
                     </TitleXLarge.Sans.Regular400>
                   </div>
--- a/src/components/UniversalNavbar/UniversalNavbar.js
+++ b/src/components/UniversalNavbar/UniversalNavbar.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import uuidv4 from 'uuid/v4'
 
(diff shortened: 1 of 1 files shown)


src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -118,13 +123,14 @@ class UniversalNavbar extends React.Component {
                   {LogoWhite({ className: 'universal-navbar-logo' })}
                 </NavLink>
                 <Spacer.H56 />
-                {LINKS.NAVLINKS.map((l) => (
-                  <div key={l.title + 'mobile'} style={{ marginBottom: 24 }}>
+                {LINKS.NAVLINKS.map((link) => (
+                  <div key={link.id} style={{ marginBottom: 24 }}>
```
roblevintennis: Booooyowwwww 👍

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -118,13 +123,14 @@ class UniversalNavbar extends React.Component {
...
                     <TitleXLarge.Sans.Regular400>
                       <NavLink
-                        key={l.title + 'mobile'}
-                        href={l.href}
-                        LinkComponent={LinkComponent}
+                        href={link.href}
+                        LinkComponent={
+                          l.href !== '/login/' ? LinkComponent : null
```
roblevintennis: Sorry, my coffee hasn't kicked in...can you explain how the `null` case would happen and result here?

iamvictorli: This is the NavLink component. 
It first checks for `LinkComponent` and since null is not truthy, we use anchor tags
```jsx
const NavLink = ({ href, LinkComponent, ...props }) => {
  if (LinkComponent) {
    return <LinkComponent to={href} {...props} />
  }
  return <a href={href} {...props} />
}
```

roblevintennis: Might be worth commenting but not a big deal and up to you. thanks for explaining

PR #64, opened 2019-10-11
Converts TextInput to use CSS Modules and removes from the compiled d…

roblevintennis:
…esign-system.css

**Description:**
- Convert TextInput to use CSS Modules. BirthdateInput and ZipInput utilize TextInput's CSS, so those too become CSS Modules compatible as well indirectly 🙌 
- Converts Error.scss to Error.module.scss--it too is only used by TextInput for inline field hint errors

- 

**Screenshots:**

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.5-c",
+  "version": "1.1.5-d",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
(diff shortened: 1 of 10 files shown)


roblevintennis:
@ethantran you should see and have some visibility on this. So, as you see, the design-system.css will continue to get trimmed down as we convert these over.

roblevintennis:
Checked corresponding CMS on chrome and my iPhone. EDS looks good too.

PR #65, opened 2019-10-11
Refactors parts of the UniversalNavbar to use css modules

roblevintennis:
**Description:**
- Trying to convert as much of UniversalNavbar to CSS Modules as feasible (sort of part 1 effot)
The FancyAnimatedLogo.scss was extracted as that's easier to leave as a global (for time being). Also TODO is maybe nuke or convert globals-for-cms.scss (not yet done here).

**Screenshots:**

image

image


src/components/UniversalNavbar/FancyAnimatedLogo.scss
```diff
@@ -0,0 +1,81 @@
+@import '../Media/Media.scss';
+
+/**
+   * Region:
+   * Fancy animated scroll behavior
+   */
+@include for-phone-only {
```
roblevintennis: I suppose if we really wanted to make this css modules, we could have the media query for every rule so it's  not nested, but I didn't want to go through all that just yet and I'm still  on the fence as to whether it's  worth it

src/components/UniversalNavbar/TransformingBurgerButton/TransformingBurgerButton.js
```diff
@@ -1,16 +1,15 @@
 import React from 'react'
+import styles from './TransformingBurgerButton.module.scss'
 
 // onClick is an event handler, showMobileMenu is a boolean
 const TransformingBurgerButton = ({ clickHandler, showMobileMenu }) => {
   return (
-    <span className="transforming-burger-button">
```
roblevintennis: Since we now have CSS Modules I no reason to keep this

src/components/UniversalNavbar/FancyAnimatedLogo.scss
tkarssli: I can't figure out why this for phone only?

tkarssli: Oh... I didn't realize that wasn't happening on any other device, was there a recent change removing the effect on other devices?

PR #66, opened 2019-10-11
Apply when on devices that support

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1137984670709817/1144506548015169/f)
- Design would  like thinner sans
- Follow up work on removing UniversalNavbar cruft
- Removes globals-for-cms.scss entirely!


src/components/design-system.css
```diff
@@ -792,213 +800,6 @@ a,
...
-.justify-around {
-  justify-content: space-around;
-}
-
-.flex-shrink-0 {
-  flex-shrink: 0;
-}
-
-/*Namespace */
-.transforming-burger-button .iconWrapper {
```
roblevintennis: This was not regenerated last merge so we're getting it here :/

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -158,7 +158,7 @@ class UniversalNavbar extends React.Component {
             <div className={styles.tabletAndUp}>
               <div className={styles.tabletAndUpContainer}>
                 {/* Desktop menu items to the left */}
-                <div className="flex items-center">
+                <div className={`${styles.flex} ${styles.itemsCenter}`}>
```
roblevintennis: Missed a  couple of globals

PR #67, opened 2019-10-12
Checkbox

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1137984670709817/1144442677379515/f)
- Checkbox. But, also adds a `useInputValidation` and removes a lot of the boiler-plate validation logic. The methods are overridable e.g. ZipInput and BirthdateInput do this


src/components/Inputs/BirthdateInput/BirthdateInput.js
```diff
@@ -66,16 +61,7 @@ const PrivateBirthdateInput = (props) => {
...
-    // User hasn't blurred but we still need to inform form
-    // engine if we're in a valid state or not
-    if (!isTouched && !!formChangeHandler) {
-      callErrorHandlers(value, formChangeHandler)
-    } else {
-      // Have blurred
-      callErrorHandlers(value, setErrorWrapper)
-    }
-  }
+  const [doValidation] = useInputValidation({validate, setError, formChangeHandler, callErrorHandlers})
```
roblevintennis: Can't see from this  diff, but I have an override `callErrorHandlers` defined before this. I also added example in the `useInputValidation` top comment, but this allows us to define custom implementations of the methods if boiler-plate of say, just one of them, doesn't work for our case.

src/components/Inputs/TextInput/TextInput.js
```diff
@@ -53,29 +50,7 @@ function PrivateTextInput({
...
-    // User hasn't blurred but we still need to inform form
-    // engine if we're in a valid state or not
-    if (!isTouched && !!formChangeHandler) {
-      callErrorHandlers(value, formChangeHandler)
-    } else {
-      // Have blurred
-      callErrorHandlers(value, setErrorWrapper)
-    }
-  }
+  const [doValidation] = useInputValidation({validate, setError, formChangeHandler})
```
roblevintennis: TextInput is an example of a component who's validation can defer to the "boiler-plate way" so no overrides. It now can just continue to call `doValidation` and all else is taken care of.

src/components/Inputs/ZipInput/ZipInput.js
```diff
@@ -20,13 +21,7 @@ export const ZipInput = (props) => {
...
   const [touched, setTouched] = useState(false)
 
-  const setErrorWrapper = (value, errorValue) => {
-    if (!!formChangeHandler) {
-      formChangeHandler(value, errorValue)
-    }
-    setError(errorValue)
-  }
-
+  // This has to come before useInputValidation setup below
```
roblevintennis: Only since it's an error function :) Another pattern would be to use a regular ole' `function` and then you wouldn't have to order these. But I really really prefer arrow functions when at all possible to use :)

src/hooks/useInputValidation.js
```diff
@@ -0,0 +1,65 @@
...
+ * @prop {function} [props.doValidation] - Optional and overrides doValidation implemented here
+ * @prop {function} [props.callErrorHandlers] - Optional and overrides callErrorHandlers implemented here
+ * @prop {function} [props.setErrorWrapper] - Optional and overrides setErrorWrapper implemented here
+ *
+ * Examples
+
+ In the simplest case to use the boiler-plate just import this hook and then do:
+   const [doValidation] = useInputValidation({validate, setError, formChangeHandler})
+
+ If you need to override doValidation, callErrorHandlers, or setErrorWrapper, do something like following:
```
roblevintennis: Hopefully speaks to my earlier comments

src/components/Form/Form.js
```diff
@@ -134,37 +134,43 @@ export function Form({ children, config }) {
       })
     }
 
+    const fieldComponent = {
```
roblevintennis: I had to refactor the field hook to only add labelCopy if the fieldConfig passed it in--CheckboxInput does not use this and I do not see why it should have to since it's children prop will have verbiage to make intent clear.

Otherwise, we'll  call back the children hook with `labelCopy` and React will error that this is not a known property for the CheckboxInput component.

src/components/Form/Form.js
```diff
@@ -134,37 +134,43 @@ export function Form({ children, config }) {
...
+      // sometimes it's passed in, but if it isn't,
+      // we will automatically generate one
+      ['data-tid']:
+        fieldConfig.tid ||
+        [config.formName, config.formId, fieldName].join('-'),
+    }
+
+    // User-visible copy, shows up in the label above the field. Most inputs
+    // have this, but CheckboxInputs do not so we only add if it applies.
+    if (fieldConfig.labelCopy) {
```
roblevintennis: I had to refactor the field hook to only add labelCopy if the fieldConfig passed it in--CheckboxInput does not use this and I do not see why it should have to since it's children prop will have verbiage to make intent clear.

Otherwise, we'll  call back the children hook with `labelCopy` and React will error that this is not a known property for the CheckboxInput component.

PR #71, opened 2019-10-14
Add classes for defining colors of type via COLORS

kenning:
Screenshot: x

These were just some missing classes. Given that this is basically a bugfix i think it's safe to just merge if approved. It's currently causing an issue in the project i'm working on (not sure how legal it is to mention it in a public repo.)


src/components/Type/Type.js
```diff
@@ -115,6 +115,7 @@ Type.COLORS = {
   // Grayscale
   GRAY_PRIMARY: COLORS.GRAY_PRIMARY,
   GRAY_SECONDARY: COLORS.GRAY_SECONDARY,
+  GRAY_STROKE_AND_DISABLED: COLORS.GRAY_STROKE_AND_DISABLED,
```
roblevintennis: Yeah, so this is fine for now and adding needed classes is also certainly fine. But  just a heads up we're going to make all this use CSS Modules soon, so all this will hopefully get rolled up into that soon.

PR #72, opened 2019-10-15
allCaps: true is the correct default

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1144809619356369/f)
- Fix the all caps issue causing QW to regress


roblevintennis:
Looks like the Form.md is showing all caps now (what we want):

image

PR #76, opened 2019-10-15
Unum password etc

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1144809619356373/f)
- Password field for Unum create page

**Screenshots:**

Invalid characters missing:

image

Invalid empty:

image

Valid:

image

### DIFF
--- a/src/components/Inputs/TextInput/TextInput.js
+++ b/src/components/Inputs/TextInput/TextInput.js
@@ -20,6 +20,7 @@ import errorStyles from '../Errors.module.scss'
  */
 
 function PrivateTextInput({
+  type,
   disabled,
   name,
   labelCopy,
@@ -27,6 +28,7 @@ function PrivateTextInput({
   formChangeHandler,
   validator,
   initialValue,
+  restrictIllegal,
   ...rest
 }) {
   // Verify that all required props were supplied
@@ -55,7 +57,7 @@ function PrivateTextInput({
 
   const onChange = (ev) => {
     const val = event.target.value
-    const restrictedVal = restrict(val)
+    const restrictedVal = restrictIllegal ? restrict(val) : val
     setValue(restrictedVal)
 
     // We call setTouched in onBlur, so can reliably call getter here
@@ -71,7 +73,7 @@ function PrivateTextInput({
 
   const onPaste = (ev) => {
     const val = ev.clipboardData.getData('text/plain')
-    const restrictedVal = restrict(val)
+    const restrictedVal = restrictIllegal ? restrict(val) : val
     setValue(restrictedVal)
   }
 
@@ -83,7 +85,7 @@ function PrivateTextInput({
     <>
       <InputLabel name={name} labelCopy={labelCopy} />
       <input
-        type="text"
+        type={type}
         className={getClasses()}
         disabled={disabled}
         name={name}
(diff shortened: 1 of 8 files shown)


src/components/Inputs/TextInput/TextInput.js
```diff
@@ -108,12 +111,18 @@ PrivateTextInput.PUBLIC_PROPS = {
   validator: PropTypes.func,
   onBlur: PropTypes.func,
   onFocus: PropTypes.func,
+  restrictIllegal: PropTypes.bool,
```
roblevintennis: We need to be able to allow special characters for password. Reality is we urlencode at the KOA side anyway so this should be fine I think.

src/validators/validatePassword.js
```diff
@@ -0,0 +1,11 @@
+const passwordRegex = /^(?=.*\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[!#$%^&*])[a-zA-Z0-9!#$%^&*]{8,100}$/
```
roblevintennis: Regex explanation:
`(?=` is a positive lookahead, and we're looking for:
- digits
- a-zA-Z (at least one upper and one lower for example)
- special characters
- 8 min - 100 max characters

Above is parity with what I saw in the legacy application, and seems to be a common approach so I just implemented it here.

ethantran: This is not allowing `@`. A legacy thing? If we are not allow that, then the default settings for most password managers like LastPass and 1Password could be rejected.

kenning: me and rob are looking around at this. but from lingke's email on 'bias for action:' 

> Sometimes the data is hard to gather, or the data itself becomes philosophically debated. In these cases, if the cost is small, and the action is easily reversible, just do it and see how it goes. Remember, mistakes are okay - especially if we can easily reverse the decision.

this is just a regex and can be updated really quickly, i'm fine with this for now and if someone complains or wants to look further into it we can investigate then.

src/validators/validatePassword.test.js
```diff
@@ -0,0 +1,32 @@
...
+  })
+
+  test('missing uppercase letter', () => {
+    expect(validatePassword('abcabc!23').length > 0).toBe(true)
+  })
+
+  test('missing lowercase letter', () => {
+    expect(validatePassword('ABCAABC!23').length > 0).toBe(true)
+  })
+})
```
kenning: Nice tests

src/validators/validatePassword.js
roblevintennis: Yeah, it's fine:

>     Make special characters more permissive with @-_ characters now allowed.

It's hard to find great data on this but:
https://discussions.agilebits.com/discussion/71544/only-certain-symbols-are-allowed-when-creating-secure-passwords

> I was just trying to change my password on the Capital One site https://www.capitalonecardservice.com/ and I ran into the following message:
> 
> "Passwords must be between 8-32 characters using letters, numbers and at least one special character @./|$*&-_"
> 

Also, looks like 1Password is kinda dynamic and has some sort of database per website??

> It is possible that they previously allowed that and now do not, or that our list of disallowed characters for that site is outdated. If the latter that should hopefully be addressed on our end in an update.

So my "bias for action" is to just add those 3 characters.

As @hilem mentioned, it's easy to update :)

roblevintennis: I should also say I appreciate the perspective you both brought to this. I elected to add in `@` and `-` in the spirit of @ethantran comment, and give up trying on the `_` in the spirit of @kenning 🤣 🤣 🤣 🤣 🤣 🤣

PR #79, opened 2019-10-16
Unum boolean buton group value

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1145052160015065/f)
- Boolean values confuse Button Select esp.  `false`

- 

**Screenshots:**

### DIFF
--- a/src/components/Form/Form.md
+++ b/src/components/Form/Form.md
@@ -287,7 +287,7 @@ const maxAge = 65
     return (
       <div>
         <TitleLarge.Serif.Book500>
-          Example Form With Birthdate
+          Example Complex Form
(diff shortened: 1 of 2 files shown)


src/components/Form/Form.md
```diff
@@ -131,6 +131,14 @@ function validateIllegal(x) {
   return x === 'illegal' ? 'That\'s an illegal option! Choose another.' : ''
 }
 
+function validateExists(x) {
```
roblevintennis: I suspect something like this should actually replace validateTruthy and we may need to do similar stuff in other fields too (not just button select groups). Maybe radio buttons, and checkboxes especially.

PR #80, opened 2019-10-16
[Don't merge] Dynamic form examples

kenning:
**Description:**
- [Asana Task](https://app.asana.com/0/1138529592199337/1145045007570209)

### I don't know whether to merge these new examples, or to essentially consider them as feature requests. This PR is mostly to open up a discussion.

This PR highlights two aspects of forms which may or may not be ideal behavior (I don't really know for sure, rob is in a meeting)

1. We don't have a way to mark a `<Form>` field as optional yet, and providing `validateExists` as the validator for a textinput is identical to not providing a validator. Example: https://i.gyazo.com/68185318e55de85df27ee3163c95d0bb.mp4 I think for now this behavior may be fine, but this could serve as documentation of the behavior for now, and could be updated later to demonstrate a `<Form>` with an optional field. ([Asana link](https://app.asana.com/0/search/1145220031676457/1145029870042133/f))
2. There is a hacky way to provide dynamic questions which you can see [here](https://github.com/getethos/ethos-design-system/pull/80/files#diff-241aaafc3b8608d9c4863b8a8fe36caeR648). Example: https://i.gyazo.com/18657b3cf05ce28b4339d0589e9c1969.mp4 As I say in that comment, it feels like a hack but it works. I tried a lot of things, you can ask for more info if you want; this was the first thing i found that behaved like we wanted.

Unfortunately prettier created a pretty messy diff. To make it easier to understand, here's a list of the other things done in this PR:

* New component used to provide dynamic state to a `<Form>` component in Form.md. https://github.com/getethos/ethos-design-system/blob/fbf528fbdd87ba8f14cfd3f3c29a168eabbf6ad9/src/components/Form/TestFormWrapper.js
* New export in `hooks/useFormState.js`, "debugEntireFormState()," which is also just used in dev. I found it useful for tracking form state and displaying hidden state in styleguidist, because styleguidist doesn't like console logs or even debugger sometimes. https://github.com/getethos/ethos-design-system/pull/80/files#diff-241aaafc3b8608d9c4863b8a8fe36caeR620

- This was a bit of a research project brought on by https://github.com/getethos/ethos/pull/1378.

### DIFF
--- a/src/components/Form/Form.md
+++ b/src/components/Form/Form.md
@@ -1,5 +1,5 @@
 ```jsx
-import validateTruthy from '../../validators/validateTruthy'
+import validateExists from '../../validators/validateExists'
 import validateMinMaxFactory from '../../validators/validateMinMax'
 import {
   TitleLarge,
@@ -49,7 +49,7 @@ const analyticsCustomEvent = (fieldName, fieldValue) => {
           return <TextInput {...props} />
         },
         validators: [
-          validateTruthy,
+          validateExists,
           validateCustom,
           validateMinMaxFactory.call(null, 5, 7),
         ],
@@ -63,7 +63,7 @@ const analyticsCustomEvent = (fieldName, fieldValue) => {
           return <TextInput {...props} />
         },
         validators: [
-          validateTruthy,
+          validateExists,
           validateCustom,
           validateMinMaxFactory.call(null, 3, 5),
         ],
@@ -138,7 +138,6 @@ const analyticsCustomEvent = (fieldName, fieldValue) => {
 _Note that we've set up the form submission to randomly fail or succeed—so, you're encouraged to resubmit until you've seen both!_
 
 ```jsx
-import validateTruthy from '../../validators/validateTruthy'
 import validateExists from '../../validators/validateExists'
 import dayjs from '../../helpers/getDayjs.js'
 import { validateMinMaxDateFactory } from '../../validators/BirthdateInputValidator'
@@ -229,7 +228,7 @@ const maxAge = 65
         // Note that Birthdate will only call these validators once it's own
         // internal validation passes e.g. there's a date string in valid format
         validators: [
-          validateTruthy,
+          validateExists,
           validateMinMaxDateFactory({
             minAge,
             maxAge,
@@ -355,9 +354,10 @@ const maxAge = 65
 ```
 
 ```jsx
-import validateTruthy from '../../validators/validateTruthy'
+import validateExists from '../../validators/validateExists'
 import {
   TitleLarge,
+  TitleSmall,
   TextInput,
   Spacer,
   Button,
@@ -374,15 +374,17 @@ import {
         component: (props, options) => {
           return <TextInput {...props} />
         },
-        validators: [validateTruthy],
-        labelCopy: 'Validator: `validateTruthy`',
+        validators: [validateExists],
+        labelCopy: 'Validator: `validateExists`',
         tid: 'example-data-tid',
       },
       noValidator: {
         component: (props, options) => {
           return <TextInput {...props} />
         },
-        labelCopy: 'No validators supplied',
+        labelCopy:
+          'No validators supplied, but you still have to focus and blur. ' +
+          'Functionally identical to the above component!',
         tid: 'example-data-tid',
       },
     },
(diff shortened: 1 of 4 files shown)


src/components/Form/Form.md
```diff
@@ -1,11 +1,18 @@
...
 import validateMinMaxFactory from '../../validators/validateMinMax'
-import { TitleLarge, TextInput, Spacer, Button, InfoMessage, ZipInput } from '../index'
+import {
+  TitleLarge,
+  TextInput,
+  Spacer,
+  Button,
+  InfoMessage,
+  ZipInput,
+} from '../index'
```
kenning: I created a pretty huge diff here just from prettier, sorry!

src/components/Form/Form.md
```diff
@@ -325,3 +353,313 @@ const maxAge = 65
...
+                  {JSON.stringify(finalFormData)}{' '}
+                </code>
+              </div>
+            </div>
+          )
+        }}
+      </Form>
+    )
+
+    // This is how the dynamic form rerenders from scratch when submitted.
```
kenning: line 648-662 here is key to understanding the current way to make forms with dynamic questions.

ethantran:
I took a glance but haven't been able to dig deeply yet. My thoughts

on optional fields:
It might be possible to make a field optional with a validator function `validateOptional(validateFunc)` where it will only run your validator function if it is not undefined. 

on form examples:
Having a final form state outside of the form is not a good idea. It is another thing to maintain. 
If I understand this correctly, it is rendering a new form on every submit? The dynamic behavior of forms should be about changing what is rendered in the current form. 
The dynamic behavior needs to be more than just based on the submit function. When I think of dynamic forms, I think of fields changing based on the state of other fields. Right now our form examples render a field with `field('name')`. How can we render it conditionally based on another field's value? "show field A when field B's value exists". "enable field C when field B's value exists"

I think the ideal implementation would look like this for the consumer of the Form component. Don't take the api literally but I think you can see where I am going with this. Whether it is done in the child render prop or in the config is debatable.
```
function DynamicExample() {
  return (
    <Form {...config}>
      {({ getFieldValue, isFieldValid, isFieldTouched, field }) => (
        <div>
          {field('A')}
          {isFieldTouched('A') && isFieldValid('A') && field('B')}
          {isFieldTouched('A') &&
            isFieldValid('A') &&
            field('C', { disabled: isFieldTouched('B') && isFieldValid('B') })}
        </div>
      )}
    </Form>
  )
}
```

roblevintennis:
@ethantran some very interesting ideas. I too don't really have much input before pulling this down. For one of your points:

> It might be possible to make a field optional with a validator function validateOptional(validateFunc) where it will only run your validator function if it is not undefined.

I'd prefer to allow for `optional` as a field config prop. You may be able to do the optional validation, and then you have several other validators...I think you'd have to somehow keep track of ordering and bail out on the other validators which I think would start to get tricky.

With `optional` configuration on a field, I consider the following rules essentially:

A field can be considered one of:
1. Valid
1. Invalid
1. Empty
Empty is really 1 or 2 and only considered when you pass in  `optional` and it means, you CAN have an empty field, but, you CANNOT have an invalid value.

I'm about to implement that bit now :)

src/components/Form/Form.md
roblevintennis: I think you pushed again and it's now 595 -- 606 lol. Yeah, I see the odd/even stuff. I'll have a look some more on train. This prototype is definitely a helpful expression of what you're after.

roblevintennis:
> The dynamic behavior of forms should be about changing what is rendered in the current form.

I think we're actually expressing opinions and two different features. Both may be needed at some point. Here's my take and I think we should call these two separate things:;

__Form Groups__
Nick needs a way to have a form represent a "question group" and be able to have a list of question groups that can run in sequence, and, each time a question group is submitted, that form needs to entirely re-render so that the fields are cleared, the submit button goes back to initial state of being disabled, etc. Running this prototype I get the use case. 

__Dynamic Forms__
Ethan would like to be able to have field conditionals based off other field conditions, and is proposing possibly adding destructured properties of the child callback `isFieldValid`, and `isFieldTouched`. I actually think that might not be too hard to support. Also, even though we have a lot of destructurable props, I think the fact that we can ask for only things we want makes it fine to add these two additions. I only wish we had a real life use case driving it—but perhaps we capture for the icebox and future iterations. I do like the idea!

So, I think you both have interesting proposals that have merit. I disagree with:
> Having a final form state outside of the form is not a good idea. It is another thing to maintain.

Well, the current forms used in single form fashion already pass the data back:
`await config.onSubmit(getFieldValues())`
and this has been working quite well from feedback so I don't actually think receiving the data is any different. What IS different is triggering the re-render, and yes, this is a challenge. Now that I understand the use case and idea, I'll be looking into this all day tomorrow and have something to show.

kenning:
Just using this instead https://github.com/getethos/ethos-design-system/pull/84

PR #81, opened 2019-10-16
Convert Layout components to CSS modules

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1143741730422930)
- This converts Layout.js and Layout.scss to use CSS modules, which removes the footgun of global CSS, as well as reducing total CSS bundle bytes by 4%.

Capture d’écran 2019-10-16 à 12 53 15

- I'll open a CMS PR that integrates these changes and some related cleanup.

**Notes:**
- Let's not merge until I've confirmed this doesn't create any issues in the CMS.


src/components/Layout/Layout.module.scss
```diff
@@ -0,0 +1,27 @@
+@import '~ethos-design-system/src/components/Media/Media.scss';
```
zach-ethos: Good catch @roblevintennis

roblevintennis: I think this is totally understandable (and took like 30 seconds to fix) because we're bouncing back and forth between the EDS and CMS so it's very easy to do this. Realistically, once we get a bit more stable on EDS these sorts of things will come up much less imo

PR #83, opened 2019-10-16
Adds the optional field marking capabilities so an optional field can…

roblevintennis:
… be left blank but the form is still considered valid for submitting

**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1145029870042133/f)
-     Optional Form Fields:

Adds the optional field marking capabilities so an optional field can be left blank but the form is still considered valid for submitting

**Screenshots:**
image

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -46,7 +46,12 @@ export function Form({ children, config }) {
 
   // Set up initial values
   let initialValues = {}
-  fieldNames.forEach((x) => {
+  fieldNames.forEach((fieldName) => {
+    console.log('in fieldNames.forEach...')
+    // We don't do the preinitialize to non empty string trick for optional
+    // fields as we'd like those to only be validated upon entering something
+    if (!!config.fields[fieldName].optional) return
+
     // By default fields have "hidden" errors declared here.
     //
     // Setting the error states to a truthy text means that the initial state
@@ -57,7 +62,7 @@ export function Form({ children, config }) {
     // If we wanted to have a form be optional, you could have its initial
     // value set to empty string, so the form could be submitted without
     // the field being touched.
-    initialValues[x] = `Initial invalid state for ${x}`
+    initialValues[fieldName] = `Initial invalid state for ${fieldName}`
   })
 
   // Hooks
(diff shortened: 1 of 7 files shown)


src/components/Form/Form.js
```diff
@@ -57,7 +62,7 @@ export function Form({ children, config }) {
     // If we wanted to have a form be optional, you could have its initial
     // value set to empty string, so the form could be submitted without
     // the field being touched.
-    initialValues[x] = `Initial invalid state for ${x}`
+    initialValues[fieldName] = `Initial invalid state for ${fieldName}`
```
roblevintennis: This is essentially what makes it all work really. @kenning just kinda going along with your thing here :)

src/components/Form/Form.js
```diff
@@ -46,7 +46,11 @@ export function Form({ children, config }) {
 
   // Set up initial values
   let initialValues = {}
-  fieldNames.forEach((x) => {
+  fieldNames.forEach((fieldName) => {
+    // We don't do the preinitialize to non empty string trick for optional
+    // fields as we'd like those to only be validated upon entering something
+    if (!!config.fields[fieldName].optional) return
```
roblevintennis: This is essentially what makes it all work really. @kenning just kinda going along with your thing here as a means to consider a non interacted with field valid. Seems to work great ;)

roblevintennis:
Following three show an `optional` with valid, invalid, and empty values:

image

image

image

src/components/Form/Form.js
```diff
@@ -169,11 +173,17 @@ export function Form({ children, config }) {
...
+    if (fieldConfig.optional) {
+      fieldComponent.optional = fieldConfig.optional
+    }
+
     return fieldConfig.component(
       fieldComponent,
 
       // For things like ButtonGroupField, which may have options supplied.
-      fieldConfig.options || null
+      fieldConfig.options || undefined
```
kenning: very minor nit but i would like it if we wrote somewhere that this repo uses undefined, not null, for null-like values or whatever you would call them. with the validateExists logic it seems like this is a pattern and it may squelch future arguments if we write it down somewhere(?)

kenning: this doesn't really need to be in this pr though

roblevintennis: Good point. Let me think about this. There's actually a ton of documentation missing/needed but just haven't got to it. But I agree this might be nice to codify somehow 👍

PR #84, opened 2019-10-17
Dynamic form examples rob

roblevintennis:
# DO NOT MERGE THIS!!
This is a PoC and removes a bunch of form examples to make it easier to focus on the experiment--so do not merge please!

Some useful things I had to "re-learn" about React's Reconcilation:

> Whenever the root elements have different types, React will tear down the old tree and build the new tree from scratch

[docs](https://reactjs.org/docs/reconciliation.html#elements-of-different-types)

To simulate this "tree tear down" it turns out you can leverage fragments and just change the key:
[keyed fragments](https://reactjs.org/docs/fragments.html#keyed-fragments)

All this is good fun and awesome, but, I'd worry if React decides to change it's API longer term if code would start to break. But it seems to me like they'll have to keep supporting this so I dunno.

The 10k foot hellicopter view takeaway to my mind is that React really really wants to apply it's diffing algorithm and not fully rerender a component if at all possible. This is sort of it's whole point (so an application developer doesn't have to manage differences between a bunch of document fragments e.g. array of: `let myFragment = document.createDocumentFragment()`. This is impractical, and so this is the big thing we get from React. 

However, for this particular type of use case, we probably DO want a full rerender (how else could we pull this off? I suppose if you managed all the fields and submit button and purposely triggered changes accordingly, you'd sort of get the same affect too and maybe that's worth investigating I dunno)

## Summary of PoC

- Is it possible to have a group of question sets all backed by one form (and thus one route / url)? 

Looks like it.

- Do we have question sets where it makes sense to not change the route and just consider them a single "suite of questions" that belong to something like /questions?

I'm not familiar enough with this particular use case. But I suppose this IS a valid use case.

- Do we want to support this in the EDS and have a `FormGroup` or `QuestionSet` or something?

My sense is while it might be worth showing off an example of how you could pull this off, it might be hard to predict what bugs could come about with this. Also, advocating for forced rerendering seems like a bad thing for a library to do.

So I'm currently 🤔 on what we should do with this and where it would need to live.


src/components/Form/Form.md
```diff
@@ -1,327 +1,184 @@
...
+  TitleLarge,
+  TitleSmall,
+  TextInput,
+  Spacer,
+  Button,
+  InfoMessage,
+  ZipInput,
+} from '../index'
+
+const questionGroups = {
```
roblevintennis: I restructured this so we could key off the question and dynamically create an array from that. I  also added `fields` as that would be closer to real use and allow for more then one field etc.

kenning:
> Still reorganizing some stuf .. but feeling Nick's pain
lol

roblevintennis:
See: https://github.com/getethos/ethos-design-system/pull/85

PR #86, opened 2019-10-18
Unum dynamic fields

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1145469163730233/f)
- Dynamic Fields


src/components/Form/Form.js
```diff
@@ -193,6 +193,8 @@ export function Form({ children, config }) {
         these arguments passed to the children function. */}
       {children({
         field,
+        errors: getFieldErrors(),
```
roblevintennis: I liked how this looked in Ethan's Formik example in terms of errors and values being property like  getterss

https://codesandbox.io/s/formik-example-tfqoi

kenning: So I sort of hoped we'd avoid a situation like this where the client sees inside the Form abstraction, because i'm a bit worried that at some point a 'form engine' starts to lose its purpose. Most of the logic is still abstracted away though, such as updating state, validation, form validity, and well, everything else you got into `<Form>`. 

I like how they're labelled with getters too. i personally dislike that `getFieldErrors` is called and passed in as `errors`. I think it's unnecessary to call the getter on every render, and the "get" prefix does a good job of implying that you're getting something effectively read-only. Would you mind passing the getter directly without changing its name? (same for `getFieldValues`)

src/components/Form/Form.md
```diff
@@ -354,6 +353,167 @@ const maxAge = 65
...
+    const {
+      field,
+      errors,
+      values,
+      getFormErrorMessage,
+      getFormIsValid,
+      getFormInteractedWith,
+    } = api
+    console.log('values', JSON.stringify(values, null, '     '))
+    console.log('errors', JSON.stringify(errors, null, '     '))
```
kenning: this is wip but yeah this and the 'form interacted with.' message should probably be removed

src/components/Inputs/TextInput/TextInput.js
```diff
@@ -90,6 +90,7 @@ function PrivateTextInput({
         className={getClasses()}
         disabled={disabled}
         name={name}
+        placeholder={rest.placeholder}
```
kenning: small but significant improvement! thanks

src/components/Form/Form.md
roblevintennis: Yeah, I removed that this morning on the way in -- haven't pushed it up yet

src/components/Form/Form.js
roblevintennis: I understand your stance on this, however, my dilemna on that is that Ethan's example has such tidy syntax that's easy to read:
```js
{values.b && !errors.b && (
                  <input...
```

Which would become:

```js
{getFieldValues().b && ! getFieldErrors().b && (
                  <input
```

And if you imagine a lot of JSX like that, it becomes unintelligible to my mind. And I'm pretty sure @ethantran would hate the later example and tell me, how come you're library is harder to use then Formik 😛 

Also, I think in fairness, we're already straying from things like getters, and  long names with `field` for the same reason—you could have called it `createField` or `newFieldFactory` or something, but it's so nice as just `field('name')` the way you did it; especially in declarative JSX land.

ethantran: I am ok with @kenning approach since the consumer can do this.
```js
{({ getFieldValues }) => {
      let values = getFieldValues()
      return {values.b ...}
}}
```

roblevintennis: Ok, if you're fine with that, then I'll use that—thanks for chiming in so quickly Ethan!

@kenning I'll do the `getFieldValues` and `getFieldErrors` thing now! :)

roblevintennis: Ok, so `{ !errors.field1 && values.field1 && field('field3')}` is interesting if you have not "form blurred" or really _form  touched_.

In terms of our current design, there's an interesting UX oddity, but it conforms to current rules.

1. `field1` has a constraint and doesn't allow more then 2 characters say
2. Upon typing first and second character, `field3` shows:

image

3. But when you type a 3rd character, `field3` will be hidden as our `!errors.field1` has happened:

image

What's perhaps surprising is that although we've taken away the field3 because we have an error, the form hasn't been touched yet, so it's technically doing the correct thing. _Note, I don't consider clicking the button group turning the form into `touched` since it just felt crappy that the error hints kick in as soon as you go into an input which is where the `touched` (blured from input) experience feels the most useful/helpful._

I'm 🤔 maybe we live with this contrived issue and see if it shows up I dunno.

src/components/Form/Form.md
```diff
@@ -354,6 +353,167 @@ const maxAge = 65
...
+        {field('buttonGroup')}
+
+        <Spacer.H16 />
+
+        {values.buttonGroup == "toggle-1" && field('field1')}
+        {values.buttonGroup == "toggle-2" && field('field2')}
+
+        <Spacer.H16 />
+
+        { !errors.field1 && values.field1 && field('field3')}
```
roblevintennis: This is the guy that's interesting due to the contrived `2 chars max allowed` on `field1`...maybe an edge case though

src/components/Inputs/TextMaskedInput/TextMaskedInput.js
```diff
@@ -84,7 +97,7 @@ export const TextMaskedInput = (props) => {
         keepCharPositions={restProps.keepCharPositions}
         pipe={restProps.pipe}
       />
-      {getError()}
+      {!doValidation && getError(currentError, formTouched)}
```
roblevintennis: `doValidation` indicates the `TextMaskedInput` consumer is doing their own validation which is how birthdate and zipcode use it. The only really want the mask rendering but not its validation (they do their own)

src/components/Form/Form.js
ethantran: The example looks good. In the future, maybe try exposing touched or interacted so `touched.x && value.x` or 'interacted.x && value.x`. See how the form behaves. Using `error` is still ok, but I think some users may find the second text input popping in and out too often just because the first input can become invalid. Although that may be just because the validator is super sensitive in your example.

ethantran: I also noticed that if you click field1 and then make first text input invalid with aaa and then click field2, you can't submit. I don't know the easiest way to fix this. That would require dynamic validation or validation of fields that depend on the state of other fields? or resetting fields when they are hidden?

roblevintennis: Yes, I saw this too. Do we need this for this week? I think I'll approach that when we get further down the line if not (or on spare cycles)

PR #90, opened 2019-10-19
Unum helptips

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1144809619356384/f)
- Help tooltips

**Screenshots:**
image

image

Also a `left` and `right` you can ask for


src/components/Tooltips/HelpTips/HelpTips.js
```diff
@@ -0,0 +1,33 @@
...
+  labelCopy,
+  helpCopy,
+  disabled,
+  position,
+  ...rest,
+}) => {
+  return (
+    <div aria-label={labelCopy} data-tid={rest['data-tid']}>
+      <span className={styles.info} data-pos={position}>i</span>
+      <span className={styles.tip} style={{ minWidth: rest['min-width'] }} >{helpCopy}</span>
```
roblevintennis: @kenning so this "works", but I'm pretty sure this isn't the final iteration/solution as it seems this would make it hard to be responsive. After I cut this code I was in the car thinking "maybe we just need to offer small / medium / large for min width, and then we can assign a `className={styles.minWidthMedium}` such that in the scss we can do our responsive media query breakpoint adjustments"...

If you feel confident in above and want to piggyback off what I have here feel free to. Otherwise, if you could advise on if you think my small/medium/large + media.scss helpers approach would work fine, lmk, and I'll take another pass.

Thanks! 🙏

kenning: I guess i'd say: having tooltips which calculate position is something people write entire js libraries for. Bootstrap for example uses "popper.js:" [link](https://getbootstrap.com/docs/4.0/components/tooltips/)

We could just use popper... It's 7.2kb https://unpkg.com/popper.js@1.16.0/dist/umd/popper.min.js

Tell me what you think about this idea

kenning: Also, I'm not 100% sure if this is what you meant by 'responsive.' Right now this is the styleguide at a thin resolution: x

roblevintennis: Yeah, so two things come to mind when narrowing the viewport:
1. min-width may need to be decreased (e.g. if we hit a media query)
1. we could just place any `left` or `right` to `above` assuming we're kinda going into mobile stacked mode I dunno.

All of these are easiest within the CSS I think, but I've got the `style={{minWidth` in the JSX which makes this harder. That sort of speaks to my original question re offering small/medium/large min width sizes.

roblevintennis: Yeah, I'm thinking about this now. Another idea (besides the small/medium/large one)...you could use `getBoundingClientRect` and somehow keep track of the container size. But I think that causes a repaint in certain browsers so I'm going to avoid it.

Ok, so, going back to idea of handling this via static CSS and only offering small/med/large. We could:

1. mobile up we could either 1. just hide the tooltips entirely (is that acceptable?) 2. reduce width, etc. and place "above" regardless of initial left/right/above/below setting
2. tablet up we maybe we don't reduce any widths, but do the _place "above" regardless of initial left/right/above/below setting_
3. desktop above we honor original settings

roblevintennis:
image

roblevintennis:
Here's what the current implementation looks like if you make `minWidth` 350 and have it centered above:

image

I suppose this might break if you having trailing [orphans](https://en.wikipedia.org/wiki/Widows_and_orphans#/media/File:Orphan-typesetting.png) and the tip is centered it'd get truncated on the left side (I think 🤔 )

roblevintennis:
More brainstorming, might be able to allow a `alignToElement` in the props. Seems like we could use `offsetLeft` on that element:

image

Will keep tinkering

kenning:
After a conversation yesterday Rob told me that this is a lost cause. Current plan is to drop monorepo's popper wrapper implementation in.

kenning:
my wip on this: https://github.com/getethos/ethos-design-system/pull/96

roblevintennis:
This was an interesting but unfruitful exercise and my take away was same as Nick's that it's probably best to revisit this and just wrap react popper as best possible. Also, growth folks might be cooking this up themselves. Closing

PR #94, opened 2019-10-21
Unum form touch issue

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1145702600849513/f)
- Form touch refactor

___

## Rationale

Looks like "field touch" or, field blur is what should determine whether field level error messages are displayed or not. 

## Research

If you look at this vanilla Formik example, we can see the default is field touch:
https://jasonwatmore.com/post/2019/04/10/react-formik-form-validation-example

Here’s an example from Jared himself (the author), and his description implies how Formik defaults to touch-based error message visibility:

> A Formik form where errors are shown while the user types (**instead of showing after blur event**). Submission is also prevented if there are any errors.

Note his example requires the application developer to override the default functionality. It also feels like the form is "yelling":

https://codesandbox.io/s/jn00j0prpv

Hence my decision to go back to field touch :)

### DIFF
--- a/src/components/Inputs/NumberInput/NumberInput.md
+++ b/src/components/Inputs/NumberInput/NumberInput.md
@@ -11,6 +11,9 @@ const formChangeHandlerStub = () => {}
   placeholder='number input'
   formChangeHandler={formChangeHandlerStub}
   validator={(n) => {
+    if (n > Number.MAX_SAFE_INTEGER) {
+      return 'Number too large—you have exceeded JavaScript\s powers!!'
+    }
     return n % 2 === 0 ? '' : 'Must be an even number'
   }}
 />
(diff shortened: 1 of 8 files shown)


src/components/Inputs/NumberInput/NumberInput.md
```diff
@@ -11,6 +11,9 @@ const formChangeHandlerStub = () => {}
   placeholder='number input'
   formChangeHandler={formChangeHandlerStub}
   validator={(n) => {
+    if (n > Number.MAX_SAFE_INTEGER) {
```
kenning: lol

roblevintennis: yeah, this was boggling me like "how come the validator stops working after a certain number of digits?" then I'm like, oh yeah, no biginty int 😋

PR #95, opened 2019-10-21
Add Email Input Validation

jameshawkinsjr:
**Description:**
- [Asana Task](https://app.asana.com/0/1116481661798430/1140084431820625)
- This is so that I can use an email input module in the CMS

- Still WIP - haven't created a PR for the email module yet

**Screenshots:**
Screen Shot 2019-10-21 at 11 41 05 AM
Screen Shot 2019-10-21 at 11 40 59 AM


src/validators/EmailValidator.js
```diff
@@ -0,0 +1,9 @@
+const validEmailRegex = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
roblevintennis: Did you write this or find it? If the later, let's leave a link back to it in a comment.

just looking at `^[<>()[\]\\.,;:\s@"]+` takes a lot of effort to mentally parse the character class and how it'd map to the first part of the email before `@` (I think). With complex regex I think documentation and links back to resources is really useful.

src/components/Inputs/EmailInput/EmailInput.test.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { EmailInput } from './index.js'
+import renderer from 'react-test-renderer'
+
+// TODO: make a helper
+// https://github.com/text-mask/text-mask/issues/427
+jest.mock('react-text-mask', () => (props) => <input type="email" {...props} />)
```
roblevintennis: Non-Blocking:
sorry about this :( hope it didn't waste too much time. I/we need to make a helper like my comment states :)

src/components/Inputs/EmailInput/EmailInput.js
```diff
@@ -0,0 +1,89 @@
...
+        type="email"
+        data-tid={restProps['data-tid']}
+        placeholder={placeholder}
+        currentValue={currentValue}
+        validator={EmailFormatValidator}
+        currentError={currentError}
+        formTouched={formTouched}
+        setFieldTouched={restProps.setFieldTouched}
+      />
+      {getError(currentError, formTouched)}
```
roblevintennis: sorry, this has changed to `touched` ... I did a big refactor this weekend and switched back to field touched (after researching Formik which uses field blurring by default). I left a lot of research, but, it means the fields pass field touched as second arg here.

src/validators/EmailValidator.test.js
```diff
@@ -0,0 +1,19 @@
+import emailFormatValidator from './EmailValidator.js'
+
+describe('EmailFormatValidator', () => {
+  it('validates 5 digit zip codes', () => {
+    // Invalid Emails
+    expect(emailFormatValidator('test_string').length === 0).toBeFalsy()
+    expect(emailFormatValidator('james@ethoslife.').length === 0).toBeFalsy()
```
roblevintennis: I think the current email validator allows james@ethoslife.z to be truthy (one character after dot). We should add a spec to confirm that still works. Probably same for `z@ethoslife.com`

jameshawkinsjr: I'll find the documentation, but the regex I used allows a minimum of 2 characters after the `.`, so  `james@ethoslife.c` wouldn't work, but ` james@ethoslife.co` works. I'll add a test in for that too.

src/validators/EmailValidator.js
jameshawkinsjr: Added a link to where I got it from - tested a few different versions and this one seemed most comprehensive.

src/components/Inputs/EmailInput/EmailInput.js
jameshawkinsjr: Updated - but am still trying to figure out if we can validation at `onBlur` rather than `onChange`

roblevintennis: You may need to pull in the latest `master` ... my sense is you don't have my changes maybe??

src/components/Inputs/EmailInput/EmailInput.test.js
jameshawkinsjr: I switched off `TextMaskedInput`, so I took this out

src/components/Inputs/EmailInput/EmailInput.js
jameshawkinsjr: I cleaned up the component quite a bit (I realized that I didn't need a lot of extra stuff that the other components were using, as this is basically a validation wrapper on top of `TextInput`. That being said, even with your changes in https://github.com/getethos/ethos-design-system/pull/94, I'm still seeing validation on `onChange`, rather than `onBlur`

src/components/Inputs/EmailInput/EmailInput.js
```diff
@@ -44,20 +25,16 @@ const PrivateEmailInput = (props) => {
...
+        disabled={disabled}
         data-tid={restProps['data-tid']}
         placeholder={placeholder}
-        currentValue={currentValue}
         validator={EmailFormatValidator}
-        currentError={currentError}
-        formTouched={formTouched}
-        setFieldTouched={restProps.setFieldTouched}
       />
-      {getError(currentError, touched)}
```
roblevintennis: Have you been able to configure it such that the TextInput is essentially kicking off the whole validation and `getError()` calls? If so, this is good in terms of removing `getError`

Are you passing through `...rest` outside of the diff? I think we need currentValue, currentError, etc.

src/components/Inputs/EmailInput/EmailInput.js
roblevintennis: I think you may need to pass in `...restProps`

src/components/Inputs/EmailInput/EmailInput.js
roblevintennis: Yeah, I think you need to let `...restProps` pass through. Not sure if that has anything to do with the fields not being validated

src/components/Inputs/EmailInput/EmailInput.md
```diff
@@ -1,5 +1,5 @@
 ```jsx
-import { EmailFormatValidator } from '../../../validators/EmailValidator'
```
roblevintennis: I actually think this was the main issue as the validator didn't exist (your exporting 'default` on other side so you can't used es6 named import here)

src/components/Inputs/EmailInput/EmailInput.md
```diff
@@ -8,14 +8,15 @@ const formChangeHandlerStub = () => {}
...
   placeholder="example@ethoslife.com"
+  validator={EmailFormatValidator}
   formChangeHandler={formChangeHandlerStub}
 />
 ```
 
 In the case where someone is returning to a form and you already have their email (i.e. editing a profile page), you can set `initialValue` which will prepopulate the form.
 
 ```jsx
-import { EmailFormatValidator } from '../../../validators/EmailValidator'
```
roblevintennis: same

src/components/Inputs/EmailInput/EmailInput.js
```diff
@@ -29,6 +29,7 @@ export const EmailInput = (props) => {
         data-tid={restProps['data-tid']}
         placeholder={placeholder}
         validator={EmailFormatValidator}
+        {...restProps}
```
roblevintennis: I'm passing these through so that the private field stuff gets passed in `currentValue`, `currentError` etc. I'm not sure if this is best idea as those would ideally stay while listed as props..but this gets it working

PR #98, opened 2019-10-22
Fix button select group initial falsy value not working (More type coercion bugs!!)

kenning:
x


src/components/Inputs/ButtonSelectGroup/ButtonSelectGroup.js
```diff
@@ -52,16 +52,19 @@ export const ButtonSelectGroup = ({
   fullWidth = true,
   ...rest
 }) => {
-  const initialSelected = currentValue || initialValue || undefined 
+  const initialSelected =
+    currentValue !== undefined ? currentValue : initialValue
```
kenning: initialValue defaults to undefined so the `initialValue || undefined` step didn't do anything. also, it would have been undefined anyway if it didn't default to undefined... haha

PR #101, opened 2019-10-22
Add 'column' property which stacks buttons

kenning:
Needed for new nicotine form: https://www.figma.com/file/iqso07OsIaQtm90DcctT5A/Unum?node-id=826%3A125


kenning:
> I like it and have been thinking we'd need to do this—thanks!

Yeah I think I remember you asking if you should do this on sunday :b

PR #102, opened 2019-10-22
Move Link to CSS modules

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1143741730422930)
- Removes Link-related globals from the CSS bundle (which shrinks 5%).
- No major issues found in CMS yet so I think this is good for final CR.
- Gonna do one last quick round of CMS QA and then I'll be ready to merge.

- https://github.com/getethos/cms/pull/1997

**Screenshots:**
- See https://github.com/getethos/cms/pull/1997.

### DIFF
--- a/src/components/Type/Link.module.scss
+++ b/src/components/Type/Link.module.scss
@@ -10,9 +10,9 @@
 
 // Style variants
 
-&.Standard {
+.Standard {
 }
 
-&.Inherit {
+.Inherit {
   color: inherit;
 }
--- a/src/components/Type/Link.module.scss
+++ b/src/components/Type/Link.module.scss
@@ -6,25 +6,13 @@
   &:hover {
     text-decoration: none;
   }
+}
 
-  // Style variants
+// Style variants
 
-  &.Standard {
-  }
+&.Standard {
+}
 
-  &.Inherit {
-    color: inherit;
-  }
-
-  &.Navlink {
-    color: var(--GrayPrimary--translucent);
-    text-decoration: none;
-    padding: var(--Space-4) 0;
-    border-bottom: 2px solid transparent;
-
-    &:active,
-    &:hover {
-      border-bottom-color: var(--GrayPrimary--translucent);
-    }
-  }
+&.Inherit {
+  color: inherit;
 }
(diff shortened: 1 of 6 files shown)


src/components/RadioButtons/RadioButtons.js
```diff
@@ -5,7 +5,7 @@ import useInvalid from '../../hooks/useInvalid.js'
 
 import { Body } from '../Type/Body.js'
 import { COLORS } from '../Colors.js'
-import styles from '../RadioButtons.module.scss'
+import styles from './RadioButtons.module.scss'
```
roblevintennis: Nick fixed this too lol...it'll FF

roblevintennis: eventually let's  node script / lint for these

src/components/Type/Link.module.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  &,
+  &:active,
+  &:hover {
+    text-decoration: none;
+  }
+
+  // Style variants
+
+  &.Standard {
```
roblevintennis: Should we consider making all these top level now that we're using CSS Modules?

zach-ethos: Yeah maybe. I went in that direction first and stopped for some reason, lemme think...

roblevintennis: Yeah, see my other comment...might be able to do this on a later iteration if it's going to hold things up. I'll let you decide but I'm fine with that if you prefer

src/components/RadioButtons/RadioButtons.js
zach-ethos: Lint PR (I wrote the code the other day, not in response to this comment, but why not open the PR?) https://github.com/getethos/ethos-design-system/pull/104

src/components/Type/Link.module.scss
zach-ethos: I'd rather keep them nested for now, unless I run into issues with it during QA. I'll update this thread a little later today.

zach-ethos:
@roblevintennis Check out the flat classes :) And we're not even using two of these styles anywhere.

PR #104, opened 2019-10-22
Eslint

zach-ethos:
Quick attempt at adding linting from the other day. This results in all files automatically being prettier-ified on each commit. This is fine but the next person the commits will get a catchup commit (can add to this PR if we want) with formatting changes.

I didn't add a preset like airbnb, standard, etc. because I think they obscure what's actually going on. More into just adding a flat list of rules we care about like @heisian does.


.eslintignore
```diff
@@ -0,0 +1,3 @@
+# These components use ES7 class statics.
+src/components/Layout/Layout.js
```
roblevintennis: https://stackoverflow.com/questions/38296761/how-to-support-es7-in-eslint

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+  extends: ['plugin:prettier/recommended'],
+  globals: {
+    Atomics: 'readonly',
+    SharedArrayBuffer: 'readonly',
+  },
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: 2018,
```
roblevintennis: I'm confused how es7 isn't supported if we're setting this. Should we be using `babel-eslint` or something? This is the type of thing I never remember and just have to research, but ideally we can lint all the files if possible. Mind investigating?

zach-ethos: Yeah I misspoke calling it ES7, it's still just a proposal (at stage 3): https://github.com/tc39/proposal-static-class-features

zach-ethos: Personally I don't see much benefit in adding a plugin so we can use an experimental syntax, when the lack of it isn't that bad (we already do a lot of post-facto statics like `Foo.propTypes = { ... }` a lot anyway for functional components). Curious what others think.

.eslintignore
zach-ethos: *Stage 3 class statics https://github.com/tc39/proposal-static-class-features

.eslintrc.js
roblevintennis: ah ok, proposal gotcha ... maybe clarify the corresponding comment then

zach-ethos:
@roblevintennis Check it out now. I took it out of the precommit until we figure out the setup and run --fix once for catchup.

.eslintrc.js
```diff
@@ -0,0 +1,49 @@
...
+    // https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
+    'plugin:prettier/recommended',
+  ],
+
+  // https://eslint.org/docs/user-guide/configuring#specifying-parser
+  // This is necessary for various ES language features we use.
+  parser: 'babel-eslint',
+
+  // https://eslint.org/docs/user-guide/configuring#specifying-parser-options
+  parserOptions: {
```
roblevintennis: Hoping these match our babel config but they look good to me

package.json
```diff
@@ -67,5 +76,13 @@
     "url-loader": "^2.1.0",
     "webpack": "^4.39.3",
     "webpack-dev-server": "^3.8.0"
+  },
+  "engines": {
+    "node": "^10.16.3"
```
roblevintennis: This turned out to break my env. I upgraded to node 10.16 and now I'm getting 

`avn could not activate node v10.16.3`

and I removed and reinstalled `avg` et al but still get this error :(

PR #108, opened 2019-10-23
[Bug fix] Fix event in onChange in TextInput

jameshawkinsjr:
`event` is incorrect in the `onChange` function in the `TextInput` component


zach-ethos:
Just ran into this, thanks :)

PR #109, opened 2019-10-23
Remove Max-width on fullWidth buttons

iamvictorli:
**Description:**
Full Width Buttons on a phone viewport is not full width when it is limited with `max-width` property

**Screenshots:**

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -64,9 +64,6 @@ button,
     padding-right: 48px;
     padding-left: 48px;
   }
-  @include for-phone-only {
-    max-width: 480px;
-  }
 }
 // Styles
 
@@ -162,7 +159,6 @@ button,
   padding: var(--Space-16) var(--Space-32);
   @include for-phone-only {
     width: 100%;
-    max-width: 400px;
   }
   @include for-tablet-and-up {
     width: auto;


src/components/Button/Button.module.scss
```diff
@@ -64,9 +64,6 @@ button,
     padding-right: 48px;
     padding-left: 48px;
   }
-  @include for-phone-only {
```
roblevintennis: Ah ok, I did https://github.com/getethos/ethos-design-system/commit/20ee3c593219c8d70bea2cbba5f56cc0fbafb9c8#diff-f8f2965423d01560e3bd62a4baabb6d1R66 because when I introduced CSS Modules I found this in CMS and removed. Are we certain it's not needed? I'm all for this if we're sure cc / @zach-ethos @kenning

src/components/Button/Button.module.scss
```diff
@@ -162,7 +159,6 @@ button,
   padding: var(--Space-16) var(--Space-32);
   @include for-phone-only {
     width: 100%;
-    max-width: 400px;
```
roblevintennis: Same for here: https://github.com/getethos/ethos-design-system/commit/6eaa934aa41eb4121bde2be7cd1c822115e4e685#diff-f8f2965423d01560e3bd62a4baabb6d1R164

cc / @kenning is this safe to remove? This was ported over from CMS for what's now `.Button.WhiteCTA`

roblevintennis:
I'm putting Zach and Nick on this as reviewers since I just moved over CMS hacks to preserve functionality

iamvictorli:
For the WhiteCTA:
Before:
Screen Shot 2019-10-23 at 11 13 34 AM
After:
Screen Shot 2019-10-23 at 11 14 18 AM

PR #110, opened 2019-10-23
Add textarea input

kenning:
Lots of js copied from TextInput, someday i would like to DRY all the internal state handling logic between our various inputs but i was happy to find that I could just import and extend the `.TextInput` class and it worked great. 

I even pulled this bad boy into the unum project and cross-scss module importing worked properly. Here's a screenshot:

x

The only other real thing going on here is it has an optional "resize" prop that defaults to false.


src/components/Inputs/TextAreaInput/TextAreaInput.md
```diff
@@ -0,0 +1,25 @@
...
+```jsx
+// formChangeHandler gets wired up automatically if using <Form /> component
+const formChangeHandlerStub = () => {}
+
+;<TextAreaInput
+  name="example"
+  labelCopy="TextAreaInput"
+  data-tid="the-textarea-input"
+  formChangeHandler={formChangeHandlerStub}
+  validator={() => ''}
```
roblevintennis: It would be nice to have a failure case here just to show the field level validation hints.

src/components/Inputs/TextAreaInput/TextAreaInput.module.scss
```diff
@@ -0,0 +1,10 @@
+// Wow so easy
+@import '../TextInput/TextInput.module.scss';
+
+.TextAreaInput {
+  @extend .TextInput;
```
roblevintennis: can we try `composes TextInput from '../TextInput/TextInput.module.scss'` ?

kenning:
composes works! just tried it in monorepo

PR #112, opened 2019-10-24
Dynamic fields problem

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1145702600849513/f)

- This PR fixes a bug @kenning found which he provided on an early commit to the Form.md file. Essentially, it's a bug for dynamic fields.
-  Dynamic fields & hidden fields—if you call  when a dynamic field is hidden, it
    will not be considered in determining form validity. Further, if the hidden field has
    been previously interacted with, it's value will also be ignored.

- It also adds some documentation in the README for the engine node requirements and how to update so `nvm` `avn` et al are happy

### DIFF
--- a/src/components/Form/BugInForm.js.example
+++ b/src/components/Form/BugInForm.js.example
@@ -1,0 +1,2 @@
+// Rename this file from BugInForm.js.example to BugInForm.js to "activate" the demo so react-styleguidist can see it
+export function BugInForm() {}
--- a/src/components/Form/Form.md
+++ b/src/components/Form/Form.md
@@ -407,7 +407,7 @@ const maxAge = 65
 </Form>
 ```
 
-Dynamic Fields Examples
+### Dynamic Fields Examples
 
 ```jsx
 import validateExists from '../../validators/validateExists'
@@ -576,6 +576,8 @@ const analyticsCustomEvent = (fieldName, fieldValue) => {
 </Form>
 ```
 
+### Hiding dynamic fields
+
 Dynamic fields & hidden fields—if you call `hide` when a dynamic field is hidden, it
 will not be considered in determining form validity. Further, if the hidden field has
 been previously interacted with, it's value will also be ignored.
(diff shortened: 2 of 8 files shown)


src/components/Form/BugInForm.js.example
```diff
@@ -0,0 +1,2 @@
+// Rename this file from BugInForm.js.example to BugInForm.js to "activate" the demo so react-styleguidist can see it
+export function BugInForm() {}
```
roblevintennis: These `BugInForm.*.example` files are meant to be a convenient template where you can quickly cut a but-topic-branch and push  up an example of the issue.

src/hooks/useFormState.js
```diff
@@ -70,7 +70,7 @@ export function useFormState(initialState) {
   // Verify form has been iteractedWith and also has no errors;
   // Used for determining whether a form is valid
   function getFormIsValid() {
-    return touched && !getFieldErrorsString()
+    return iteractedWith && !getFieldErrorsString()
```
roblevintennis: This is the bugfix for https://app.asana.com/0/1136962714401929/1145702600849513/f

src/components/Form/Form.md
```diff
@@ -1,3 +1,106 @@
...
+;<Form
+  config={{
+    formName: 'Dynamic fields example form',
+    autocompleteOff: true,
+    formId: '1',
+    fields: {
+      buttonGroup2: {
+        component: (props, options) => {
+          return (
+            <ButtonSelectGroup {...props} intialValue="often">
```
jameshawkinsjr: I can't get `intialValue="____"` to do anything here

src/components/Form/Form.md
```diff
@@ -1,3 +1,106 @@
+### Dynamic field issue
```
jameshawkinsjr: When interacting with this form, my assumption would be that if I selected `I refuse, please let me submit`, it would allow me to submit the form, but that doesn't seem to be the case.

src/hooks/useFormState.js
```diff
@@ -70,7 +92,17 @@ export function useFormState(initialState) {
...
+  }
+
+  function getFieldsHidden() {
```
kenning: This is never called anywhere.

src/hooks/useFormState.js
```diff
@@ -2,12 +2,21 @@ import { useState } from 'react'
 
 let iteractedWith = false
```
kenning: This is still misspelled :p

src/hooks/useFormState.js
roblevintennis: removed

src/hooks/useFormState.js
roblevintennis: fixed

roblevintennis:
I've addressed the issues Nick found and he's approved so merging this.

roblevintennis:
Regarding the side effects and render I’m curious how calling high does any different than calling field? At the core of how this works is within the field method we set hidden to false and hide method true. I don’t use state which triggers rerender but a local variable and so rerendering isn’t triggered

PR #113, opened 2019-10-25
Debloat tmp off dynamic fields problem

roblevintennis:
debloat (I cut this off of `dynamic-fields-problem` which will go in tomorrow so no worries):

image


src/components/Form/Debloat.js
```diff
@@ -0,0 +1 @@
+export function Debloat() {}
```
roblevintennis: We have to trick react-styleguidist to render the component and so we have to put this here

src/components/Form/DebloatStub.js
```diff
@@ -0,0 +1,46 @@
...
+import EmailFormatValidator from '../../validators/EmailValidator'
+
+// This might be plucked out of routeData.data.emailpage.fields or something similar
+const json = {
+  email: {
+    placeholder: 'example@ethoslife.com',
+    name: 'the-email',
+    labelCopy: 'Email ',
+    tid: 'the-email-tid',
+    validators: ['EmailFormatValidator'],
```
roblevintennis: turns out to be easy to map these one to one if we simply use a lookup table. Also, if we change the names we just do:

```js
const validatorsTable = {
  validatePassword: IDecidedToMakeItWeirdAndRenameValidatePasswordButIStillWork
...
```

src/components/Inputs/PasswordInput/index.js
```diff
@@ -1 +1 @@
-export { TextInput } from './TextInput.js'
+export { PasswordInput } from './PasswordInput.js'
```
roblevintennis: This is actually a bugfix!

src/components/Form/Debloat.md
```diff
@@ -0,0 +1,82 @@
...
+  InfoMessage,
+  ZipInput,
+  EmailInput,
+  Form,
+  PasswordInput,
+} from '../index'
+
+let count = 1;
+
+;<Form
```
roblevintennis: the leading ; is just an annoyance we have to do for react-styleguidist so it doesn't get confused

src/components/Form/DebloatStub.js
```diff
@@ -0,0 +1,46 @@
...
+    validators: ['validatePassword'],
+  },
+}
+
+const validatorsTable = {
+  validatePassword: validatePassword,
+  EmailFormatValidator: EmailFormatValidator,
+}
+
+// Assumes json has been parsed already
```
roblevintennis: meant parsed into a json object literal

src/components/Form/DebloatStub.js
```diff
@@ -0,0 +1,57 @@
...
+const mapJsonToFields = (json) => {
+  const fields = {}
+  Object.keys(json).forEach((fieldKey) => {
+    fields[fieldKey] = {}
+    const field = json[fieldKey]
+    fields[fieldKey] = field
+    const validators = field.validators.reduce((acc, validator) => {
```
alexthecarp: nit: this would also work i think:` const validators = field.validators.map((validator) => validatorsTable[validator])`

alexthecarp:
LGTM! Approved, 1 nit.

src/components/Form/DebloatStub.js
roblevintennis: done in 71323c8101ca2d8838f574d19d891d3b7a9e07af

src/components/Form/Resolver.js
```diff
@@ -0,0 +1,17 @@
+// Assumes json has been parsed into a JavaScript object literal
+const mapJsonToFields = (json, componentsLookup, validatorsLookup) => {
```
roblevintennis: @alexthecarp There is really no meaningful change from the last version you saw, but, this codifies what was _in my head_ in terms of having a "Resolver" that maps somewhat arbitrary json, component tables, and validator tables, but in a way that should allow for various per use case setups. The team can set these up pretty much however it makes sense for whatever form they need to support.

roblevintennis:
I'm going to close this since we've incorporated it into our debloat POC: https://github.com/getethos/ethos/pull/1472

PR #114, opened 2019-10-25
Downgrading

roblevintennis:
**Description:**
The node engine requirement would require monorepo et al to upgrade node. It seems like eslint runs fine with 10.6 so this PR downgrades to match what we have across our stack.

- https://github.com/getethos/ethos-design-system/pull/104/files

`yarn lint` runs fine. We have a lot to fix apparently, but that should be done on a follow up branch I think since it's not blocking our consumers as the node engine thing IS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,9 +79,6 @@
     "webpack": "^4.39.3",
     "webpack-dev-server": "^3.8.0"
   },
-  "engines": {
-    "node": "^10.16.3"
-  },
   "husky": {
     "hooks": {
       "pre-push": "yarn test"


package.json
```diff
@@ -80,7 +80,7 @@
     "webpack-dev-server": "^3.8.0"
   },
   "engines": {
-    "node": "^10.16.3"
+    "node": "^10.6.0"
```
roblevintennis: I can tell this is working by CD'ing out of the EDS and back into it, and `avn` sets the node version to 10.6 as expected.

roblevintennis:
image

roblevintennis:
__Update__

Just removing the `engines` clause seems to be best move as it allows us to keep the new stuff we want mainly for our own dev linting etc., but consumers aren't forced to upgrade node.

PR #115, opened 2019-10-25
Revert "Eslint (#104)"

zach-ethos:
This reverts commit ae81997fd11a14af453a0e995512fbbd7d9bcc14.


roblevintennis:
I'm going to try just removing the engine clause. We should be able  to use whatever version of node et al we want, just not force consumers to since we're don't have many

PR #117, opened 2019-10-26
hide becomes markHidden

roblevintennis:
## This will require any consumers that use `hide` to refactor to use `markHidden`

**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146709581895113/f)
-  hide becomes markHidden since we do not hide the thing but mark it as hidden thus do not consider when returning for values or validating fields for errors etc.

**Screenshots:**

This is a temporary page I created...but I marked buttonGroup 2 and 3 as hidden after clicking the 'Yo' button...the console.log shows that we only returned button 1 in the  form state collection

image

### DIFF
--- a/src/components/Form/Form.js
+++ b/src/components/Form/Form.js
@@ -119,9 +119,10 @@ export function Form({ children, config }) {
     }
   }
 
-  function hide(fieldName) {
+  function markHidden(fieldName) {
     const fieldConfig = config.fields[fieldName]
     setFieldsHidden(fieldName, true)
+    return true
   }
 
   // Wrapper for all fields. Essentially, this translates the field definitions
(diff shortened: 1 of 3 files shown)


src/components/Form/Form.js
```diff
@@ -119,9 +119,10 @@ export function Form({ children, config }) {
     }
   }
 
-  function hide(fieldName) {
+  function markHidden(fieldName) {
     const fieldConfig = config.fields[fieldName]
     setFieldsHidden(fieldName, true)
+    return true
```
roblevintennis: allows for chaining if needed

src/components/Form/Form.js
```diff
@@ -122,6 +122,7 @@ export function Form({ children, config }) {
   function markHidden(fieldName) {
     const fieldConfig = config.fields[fieldName]
     setFieldsHidden(fieldName, true)
+    // allows for chaining with && if needed
```
roblevintennis: 👍

PR #118, opened 2019-10-26
Prefill still requires touching fields

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146473861245233/f)
- Questions Page - Continue disabled with all inputs prefilled

Essentially, `initialValue` for a given field will now result:

1. field is considered _touched_
2. the field's validation will get called. This means the form state is a aware of the field's validity

If the initial values are valid values and all for fields are using this, the `buttonIsValid()` will return true and thus the submit button will show as enabled.

**Screenshots:**

image

** Notes: **

To verify interact with http://localhost:9008/#/Components/FormPrefilled and try manually setting the `initialValue` for the fields to invalid values and watch that the submit button gets disabled.

### DIFF
--- a/src/components/Inputs/BirthdateInput/BirthdateInput.js
+++ b/src/components/Inputs/BirthdateInput/BirthdateInput.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { TextMaskedInput } from '../TextMaskedInput'
 import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe'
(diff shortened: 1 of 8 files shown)


src/components/Inputs/BirthdateInput/BirthdateInput.js
```diff
@@ -69,6 +69,15 @@ const PrivateBirthdateInput = (props) => {
     }
   }
 
+  // Initial Value aka prefilled—are considered "touched", but must prevalidate
+  // which will in turn update the internal form state as to their validity
+  useEffect(() => {
+    if (!!formChangeHandler && initialValue) {
+      console.log('BirthdateInput useEffect -- calling doValidation')
```
roblevintennis: I'm going to leave these console.log's for now until we've validated this works in the Consumer itself (the secret project we're working on that may not be named publicly 🤣 )?

The big idea here is since we're passing in `[]` this effect should only run once on first render (as we want), and so the initial value will be considered as we require.

PR #123, opened 2019-10-28
Test coverage

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146729483498353/f)
- Follow up EDS Tasks:
    - Form Test Coverage


roblevintennis:
@kenning  you can do `yarn e2e:dev`

src/components/Inputs/CheckboxInput/CheckboxInput.js
```diff
@@ -75,7 +77,9 @@ export const CheckboxInput = ({
         ev.target.type === 'checkbox' ? ev.target.checked : ev.target.value
       doValidation(!val, touched)
       setIsChecked(!val)
-      setFieldTouched(true)
+      if (!!setFieldTouched) {
```
roblevintennis: This only gets called when we're hitting the field from Form.js e.g. if we use it "directly" this will likely not be defined

package.json
```diff
@@ -8,10 +8,10 @@
     "build": "yarn compile-css",
     "compile-css": "node-sass src/components/index.scss src/components/design-system.css && yarn format",
     "cy:open": "cypress open",
-    "cy:run": "cypress run --browser chrome --record",
+    "cy:run": "cypress run --browser chrome --config video=false",
     "develop": "chokidar src/**/*.scss -c 'yarn build' --initial",
-    "e2e:ci": "CYPRESS_SUPPORT=y yarn run build && start-server-and-test styleguide http://localhost:9008 cy:open",
-    "e2e:dev": "CYPRESS_SUPPORT=y start-server-and-test styleguide http://localhost:9008 cy:open",
+    "e2e:ci": "export NODE_ENV=ci && start-server-and-test styleguide http://localhost:9008 cy:run",
```
iamvictorli: Travis supports Env variables in their dashboard, so there is no need to add `NODE_ENV=ci`

.travis.yml
```diff
@@ -1,7 +1,17 @@
 language: node_js
-install:
-  - yarn
+node_js:
+  - 'lts/*'
```
iamvictorli: Since EDS runs on node v10.16.3, we probably shouldn't be using cutting edge `lts/*`
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc
If these are deleted, Travis would read `.nvmrc` file and know to use a node version v10.16.3

package.json
roblevintennis: I've actually moved this over into the yml config which feels correct.

.travis.yml
roblevintennis: fixed

package.json
iamvictorli: That works as well. If you ever wanted to put sensitive env variables, the way to put it is to go to settings -> environment variables
Screen Shot 2019-10-28 at 8 49 48 PM

roblevintennis: Well, I understand that you can actually [encrypt those too](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml) and still put in the `env` stanza, but we just needed a boolean to indicate CI.

iamvictorli: Wow amazing. TIL

.travis.yml
```diff
@@ -1,7 +1,17 @@
...
+      - libgconf-2-4
+install: yarn install --frozen-lockfile
+cache:
+  yarn: true
+  directories:
+    - ~/.cache
+env:
+  - NODE_ENV=ci
 script:
   - yarn test
```
iamvictorli: Travis should also run a linting step before `testing`

roblevintennis: We have https://app.asana.com/0/1136962714401929/1146342193058673/f is to fix our linting errors, and then also add linting before test as you're recommending, thanks, but I don't want to wait to get this in before hooking up CI/CD as ensuring the Form Engine API doesn't get inadvertently borked is top priority for all this to me, and I'd like to get that defense in if possible asap.

cypress/integration/FormPrefilled.js
```diff
@@ -0,0 +1,17 @@
+describe('FormPrefilled', () => {
+  beforeEach(() => {
+    cy.visit('/#/Components/FormPrefilled')
+  })
+
+  it('fully prefilled form should result in submit button enabled', () => {
+    cy.getByTid('button-prefilled-submit').should('not.be.disabled')
```
jameshawkinsjr: If you change this line to `cy.getByTid('button-prefilled-submit').should('not.be.disabled')` the tests still pass - seems like this isn't doing what's asserted.

roblevintennis: Thanks, I'll have a look

iamvictorli:
These should be added to `.gitignore`
cypress/videos
cypress/screenshots

https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Configuring-Folder-Structure
>Cypress will create a screenshotsFolder and a videosFolder to store the screenshots and videos taken during the testing of your application. Many users will opt to add these folders to their .gitignore file.

zach-ethos:
I have the same issue as @jameshawkinsjr, but otherwise it runs and LGTM. I tried `'have.attr'` for a bit with similar issues.

cypress/integration/FormPrefilled.js
roblevintennis: @iamvictorli @jameshawkinsjr Yeah, this is super annoying. This is sort of the most important assertion in the spec saying "if I've filled all fields with valid `initialValue`'s, the submit should not be disabled". Open to any suggestions or workarounds, but I agree I can't quite merge this unfortunately until it gets figured out :/

roblevintennis: Ok, I just realized, if you look at this in the styleguide for real and look carefully and reload the page scrolled to the Submit button, it turns out the button starts disabled. So, although I usually hate using `wait`, I think this use case warrants it. I tried lower settings of 50 ms, but that's too fragile and 200ms seems to work well. I tried it 10 times in a row with `be.disabled` and it failed 10x. I also toggled back and forth between that and `.not.be.disabled` and got expected results.

Here's it failing for example:
image

roblevintennis:
@zach-ethos 

image


Just ran `yarn lint` after merging in `master`. Pretty sure a non-zero exit code will fail all builds. I do want to get those errors down to 0, but I'm considering commenting it out and then enabling it once we do. I'm a bit concerned to get coverage in CI/CD for especially the FormPrefilled and dynamic fields functionality if possible; we need to at least some defense against regressions there.

roblevintennis:
Yeah, it does fail in Travis for this as I suspected it would:

image

I'ma try commenting out the line so it's sitting there staring at us to put back once we get the lint errors down.

.travis.yml
```diff
@@ -1,11 +1,23 @@
...
+    - ~/.cache
+env:
+  - NODE_ENV=ci
 
 install:
   - yarn
 
 script:
-  - yarn lint
+  # - yarn lint
```
roblevintennis: This will get added back in once we're down to zero lint errors ... as it stands it'll return a nonzero atm. Unfortunate yes, but I don't want to hold up getting coverage for the Form stuff especially.

roblevintennis:
@jameshawkinsjr @ethantran @iamvictorli @zach-ethos can I get a final review on this please? Thanks 🙏

iamvictorli:
Do you mind adding these to `.gitignore`
- cypress/videos
- cypress/screenshots
These shouldn't be in version control

https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Configuring-Folder-Structure
>Cypress will create a screenshotsFolder and a videosFolder to store the screenshots and videos taken during the testing of your application. Many users will opt to add these folders to their .gitignore file.

roblevintennis:
@jameshawkinsjr no thanks for the 411! I used the same fix for: ec0006b2243bd7cc42bbd01b3cdd8c23d2b2d466 ...

I added the `cy.wait` and was able to make this fail 10x in a row:
https://github.com/getethos/ethos-design-system/blob/master/cypress/integration/FormDynamic.js#L48

image

___

Same for #89 example. Ran 10x times with inverted assertion and it always failed:
image

.travis.yml
zach-ethos: Even with --max-warnings?

roblevintennis: It did yesterday. Maybe since we've merged some other stuff it won't?? I dunno, the `--max-warnings` was definitely there and it failed. It hasn't been of late locally so maybe it'll work.

My hope is to get the lint errors zero very soon and enable anyway though

PR #125, opened 2019-10-30
Add aria-hidden, tests to Spacer (a11y fix)

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1128987846241693/1117917027835300)
- Ideally we do spacing in CSS. To the extent that we use HTML we should be adding aria-hidden to hide the nonsemantic tags from screen readers.
- I added this elsewhere and realized we should add it here. It's a one-liner + tests.


roblevintennis:
I'm confused, all . the tests work for me locally, but CI is failing on some `aria` stuff in the snapshots

PR #126, opened 2019-10-30
Fix lint problems (round 1)

zach-ethos:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146342193058673)
- Earlier we had ~140 lint errors. This fixes a handful. I'm trying to break this into small PRs.
- @roblevintennis take a glance at the ternaries at the end? I think they're fine but on an earlier run I thought I saw them get mangled. (See also: https://prettier.io/docs/en/cli.html#debug-check)

### DIFF
--- a/src/components/Inputs/TextMaskedInput/TextMaskedInput.js
+++ b/src/components/Inputs/TextMaskedInput/TextMaskedInput.js
@@ -60,7 +60,7 @@ export const TextMaskedInput = (props) => {
 
   const setAllTouched = () => {
     whichSetTouched(true)
-    if (!!setFieldTouched) {
+    if (setFieldTouched) {
       setFieldTouched(true)
     }
   }
@@ -93,7 +93,7 @@ export const TextMaskedInput = (props) => {
   }
 
   const getClasses = () => {
-    return !!getError(currentError, whichTouched)
+    return getError(currentError, whichTouched)
       ? `TextMaskedInput ${styles.TextInput} ${errorStyles.Error}`
       : `TextMaskedInput ${styles.TextInput}`
   }
--- a/src/hooks/useInputValidation.js
+++ b/src/hooks/useInputValidation.js
@@ -33,18 +33,22 @@ const useInputValidation = ({
   callErrorHandlers,
   setErrorWrapper,
 }) => {
-  const setErrorWrapperInternal = setErrorWrapper ? setErrorWrapper : (value, errorValue) => {
-    if (!!formChangeHandler) {
-      formChangeHandler(value, errorValue)
-    }
-    setError(errorValue)
-  }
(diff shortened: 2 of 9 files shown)


src/components/Inputs/TextMaskedInput/TextMaskedInput.js
```diff
@@ -60,7 +60,7 @@ export const TextMaskedInput = (props) => {
 
   const setAllTouched = () => {
     whichSetTouched(true)
-    if (!!setFieldTouched) {
+    if (setFieldTouched) {
```
roblevintennis: I think we actually need this since you only get this if you hit these from Form.js unfortunately. So we need to preserve all `!!setFieldTouched` parts

src/hooks/useInputValidation.js
```diff
@@ -33,18 +33,22 @@ const useInputValidation = ({
...
+        setError(errorValue)
+      }
 
-  const callErrorHandlersInternal = callErrorHandlers ? callErrorHandlers : (value, handlerFn) => {
-    let errorMessage = validate(value)
-    errorMessage = errorMessage.length ? errorMessage : ''
-    handlerFn(value, errorMessage)
-  }
+  const callErrorHandlersInternal = callErrorHandlers
+    ? callErrorHandlers
```
roblevintennis: This doesn't seem to bad to me

src/components/Inputs/TextMaskedInput/TextMaskedInput.js
```diff
@@ -93,7 +93,7 @@ export const TextMaskedInput = (props) => {
   }
 
   const getClasses = () => {
-    return !!getError(currentError, whichTouched)
```
roblevintennis: Can we just configure eslint to allow this? I think it's a bit nuanced when this is good or bad.

src/components/Inputs/TextMaskedInput/TextMaskedInput.js
zach-ethos: I'm not sure I understand. How does the `!!` here matter?

src/components/Inputs/TextMaskedInput/TextMaskedInput.js
zach-ethos: This seems more problematic. I.e., if you were expecting a boolean later and now you don't have one. Lemme see how to disable this.

src/components/Inputs/TextMaskedInput/TextMaskedInput.js
roblevintennis: I was wrong on this. I just pulled `fix/lint-problems-i` and it's fine 👍

PR #127, opened 2019-10-30
Add ability to hide cta on desktop to navbar

kenning:
necessary for https://app.asana.com/0/1116481661798430/1136005494202191

### DIFF
--- a/src/components/UniversalNavbar/UniversalNavbar.js
+++ b/src/components/UniversalNavbar/UniversalNavbar.js
@@ -8,7 +8,7 @@ import LogoWhite from './assets/ethos-logo-white.js'
 import { Button, Layout, Spacer, TitleXLarge } from '../index'
 import TransformingBurgerButton from './TransformingBurgerButton/TransformingBurgerButton'
 
-import styles from  './UniversalNavbar.module.scss'
+import styles from './UniversalNavbar.module.scss'
 
 // TODO REDESIGN: Lots of sloppy inline styles here.
 // TODO: Remove last usages of the Media helper (and prefer the Sass MQ mixins).
@@ -67,7 +67,7 @@ class UniversalNavbar extends React.Component {
   }
 
   render() {
-    const { LinkComponent, hideMobileCta } = this.props
+    const { LinkComponent, hideMobileCta, hideDesktopCta } = this.props
 
     const getAnEstimate = (showWhenScrolled) => (
       <a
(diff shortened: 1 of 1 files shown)


src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -194,15 +194,18 @@ class UniversalNavbar extends React.Component {
 
 UniversalNavbar.propTypes = {
   /** Hide cta on mobile viewport */
-  hideMobileCta: PropTypes.bool.isRequired,
+  hideMobileCta: PropTypes.bool,
+  /** Hide cta on desktop */
+  hideDesktopCta: PropTypes.bool,
   /** Run analytics function when CTA Button gets clicked */
-  trackCtaClick: PropTypes.func.isRequired,
+  trackCtaClick: PropTypes.func,
```
iamvictorli: I think this should be required, as long as defaultProps is defined.
From the site [React doc site](https://reactjs.org/docs/typechecking-with-proptypes.html):
> The propTypes typechecking happens after defaultProps are resolved, so typechecking will also apply to the defaultProps.

kenning: what's the point of marking them required if they have defaults?

kenning: i'll just do it but i am actually curious what you think about this

iamvictorli: You have a point. Never thought about that. There was this ESLint rule I used before, where everything had to be required or be in defaultProps. You have a point.

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -203,6 +205,7 @@ UniversalNavbar.propTypes = {
 
 UniversalNavbar.defaultProps = {
   hideMobileCta: false,
+  hideDesktopCta: false,
```
roblevintennis: I agree with you Nick, let's just remove this if it's required anyway

src/components/UniversalNavbar/UniversalNavbar.js
roblevintennis: > what's the point of marking them required if they have defaults?


Inversely, there's no reason to add a default if  something is required. For example:
[this eslint](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md) states:

> Enforce a defaultProps definition for every prop **that is not a required prop** (react/require-default-props)

So I'd like us to remove the default values for required it's just adding more noise.

PR #128, opened 2019-10-30
Fix lint problems (round 2)

zach-ethos:
**Description:**
- [Round 1](https://github.com/getethos/ethos-design-system/pull/126) 
- Mostly straightforward. I'll tag a few people as a heads up.

PR #130, opened 2019-10-30
Cap lint problems

zach-ethos:
**Description:**
- [Round 1](https://github.com/getethos/ethos-design-system/pull/126), [Round 2](https://github.com/getethos/ethos-design-system/pull/128)  
- This should prevent further lint errors by failing locally and in CI if you add more.
- The basic idea is, we have 82 issues, and if that goes up, the build fails. Later we reduce 82 to 0.
- TLDR: this doesn't fix everything we could possibly fix, but it prevents things from worsening.

**Notes:**

- The biggest source of remaining lint problems (30%) is propTypes in Form.js.
- The rest are propTypes elsewhere and `no-unused-vars`.
- The docs for --max-warnings: https://eslint.org/docs/user-guide/command-line-interface#exit-codes

**Screenshots:**

Here is what it looks like when you add a new lint problem (unused variable `foo`):

Capture d’écran 2019-10-30 à 16 14 19

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "e2e:ci": "CYPRESS_SUPPORT=y yarn run build && start-server-and-test styleguide http://localhost:9008 cy:open",
     "e2e:dev": "CYPRESS_SUPPORT=y start-server-and-test styleguide http://localhost:9008 cy:open",
     "format": "prettier --write \"src/components/**/*.*css\"",
-    "lint": "eslint src",
+    "lint": "eslint src --max-warnings 82",
     "prerelease": "yarn test && yarn build",
     "spotcheck": "yarn compile-css && echo \"\n*****\n* View the spotcheck at http://localhost:8080/spotcheck.html\n*****\n\" && webpack-dev-server ",
     "stats": "node bin/generateStats.js --write",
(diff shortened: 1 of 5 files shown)


zach-ethos:
Here's how it looks on Travis:

https://travis-ci.com/getethos/ethos-design-system/builds/134296169#L334

Capture d’écran 2019-10-30 à 16 28 46

src/components/Spacer/Spacer.test.js
```diff
@@ -95,35 +92,6 @@ describe('Spacer', () => {
...
-      expect(Spacer.W56).toBeDefined()
-      expect(Spacer.W48).toBeDefined()
-      expect(Spacer.W40).toBeDefined()
-      expect(Spacer.W32).toBeDefined()
-      expect(Spacer.W24).toBeDefined()
-      expect(Spacer.W16).toBeDefined()
-      expect(Spacer.W8).toBeDefined()
-      expect(Spacer.W4).toBeDefined()
-    })
-  })
```
zach-ethos: I think these tests are redundant now that we have snapshots.

package.json
```diff
@@ -13,7 +13,7 @@
     "e2e:ci": "CYPRESS_SUPPORT=y yarn run build && start-server-and-test styleguide http://localhost:9008 cy:open",
     "e2e:dev": "CYPRESS_SUPPORT=y start-server-and-test styleguide http://localhost:9008 cy:open",
     "format": "prettier --write \"src/components/**/*.*css\"",
-    "lint": "eslint src",
+    "lint": "eslint src --max-warnings 82",
```
roblevintennis: That's a pretty cool rule 👍

PR #137, opened 2019-10-31
Add ButtonSelectGroup propTypes

zach-ethos:
This plus the pending branches take us down to ~30 issues.


roblevintennis:
@zach-ethos I think this may have caused some invalid prop types:
image


I'm going to look into it now and push up a fix.

PR #140, opened 2019-11-01
Fix more issues found with PropTypes

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146342193058673/f)
- Fixes rest of prop type warning besides Form.js (once https://github.com/getethos/ethos-design-system/pull/139 is also in)

### DIFF
--- a/src/components/Inputs/EmailInput/EmailInput.js
+++ b/src/components/Inputs/EmailInput/EmailInput.js
@@ -27,6 +27,7 @@ export const EmailInput = (props) => {
         optional={optional}
         type="email"
         data-tid={restProps['data-tid']}
+        disabled={disabled}
         placeholder={placeholder}
         validator={EmailFormatValidator}
         {...restProps}
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -48,7 +48,7 @@ class ScrollDetector extends React.Component {
   }
 
   render() {
-    const { className, element, ...rest } = this.props
+    const { className, element, offsetHeight, ...rest } = this.props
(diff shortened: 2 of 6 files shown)


src/components/Layout/Layout.js
```diff
@@ -48,7 +48,7 @@ class ScrollDetector extends React.Component {
   }
 
   render() {
-    const { className, element, ...rest } = this.props
+    const { className, element, offsetHeight, ...rest } = this.props
```
roblevintennis: @iamvictorli @tkarssli @zach-ethos guys, I had to do this to prevent:

image

Which seems to reflect [transferring props](https://reactjs.org/warnings/unknown-prop.html) ... I believe this is fine since it seems we only need to use this in the component itself for:

```js

  updateScrollState = () => {
    const scrollTop = window.pageYOffset || window.document.scrollTop
    this.setState({ isScrolled: scrollTop > this.props.offsetHeight })
  }
```

Just a heads up though!

src/components/Inputs/EmailInput/EmailInput.js
```diff
@@ -27,6 +27,7 @@ export const EmailInput = (props) => {
         optional={optional}
         type="email"
         data-tid={restProps['data-tid']}
+        disabled={disabled}
```
roblevintennis: This was just missing so we couldn't pass through disabling of element down to the TextInput email delegates too.

src/components/Layout/Layout.js
iamvictorli: This is an issue that we see on CMS. Thanks for looking at this

PR #145, opened 2019-11-05
Radio buttons group

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1143351621986088/f)
- 

Form engine demo page:

image

Radio button page:

image


src/components/RadioButtons/RadioButtons.js
```diff
@@ -1,63 +1,17 @@
...
+import uuidv4 from 'uuid/v4'
+import useErrorMessage from '../../hooks/useErrorMessage.js'
 import useIncludes from '../../hooks/useIncludes.js'
 import useInvalid from '../../hooks/useInvalid.js'
 
 import { Body } from '../Type/Body.js'
 import { COLORS } from '../Colors.js'
 import styles from './RadioButtons.module.scss'
 
-// Note: RadioButtonGroup does not currently work (may have broken circa v0.9).
```
roblevintennis: I'm opting to completely remove the docs since they're from a previous version and I'd rather we rebuild these up thoughtfully as out of date docs will do more harm then good imo

src/components/RadioButtons/RadioButtons.js
```diff
@@ -98,47 +52,72 @@ RadioButton.propTypes = {
...
-  // also disable an individual option/radio button via `option.disabled`.
-  // Not supported yet: autoComplete, tabIndex (not obviously necessary)
+  let initialSelected
+  if (currentValue || typeof currentValue === 'boolean') {
+    initialSelected = currentValue
+  } else if (initialValue || typeof initialValue === 'boolean') {
+    initialSelected = initialValue
+  }
+
+  const [selectedValue, setSelectedValue] = useState(initialSelected)
```
roblevintennis: There's some known duplication between this and ButtonSelectGroup. I don't want to overdo it on this PR, so I've created a follow up task for DRY'ing that up:

https://app.asana.com/0/1137984670709817/1148024641301841/f

roblevintennis:
Here's debloat pointing to this branch (from `update-frontend-backend-poc` ethos monorepo branch):

image

roblevintennis:
@alexthecarp @ekeric13 cc / this is the branch I've pointed us to for the _Intent_ question

src/components/RadioButtons/RadioButtons.js
kenning: hard to disagree with

src/components/RadioButtons/RadioButtons.js
kenning: 🙌

src/components/RadioButtons/RadioButtons.js
```diff
@@ -69,10 +33,11 @@ function RadioButton({
           name={name}
           value={value}
           checked={checked}
+          onClick={onClick}
+          onChange={onChange}
           required={required}
           disabled={disabled}
           data-tid={rest['data-tid']}
-          {...rest} // for redux-form input.onChange, etc.
```
kenning: 🙌

src/components/RadioButtons/RadioButtons.js
```diff
@@ -98,64 +64,96 @@ RadioButton.propTypes = {
...
- * @param {String}     value     The currently selected option.value
- * @param {Boolean}    disabled  Whether this radio group is disabled
- * @param {Boolean}    required  Whether this radio group is required
- * @param {...Object}  rest      Mostly redux-form Field input/meta props
- */
 export function RadioButtonGroup({
-  name,
   options,
-  value,
+  onChange,
```
kenning: Other inputs do not have a public onchange handler, i don't think this one should have one. It gets called below in useEffect, and is also supplied in the form.md example.

kenning:
Made a couple small changes that would be more work to request than to just do. I made it so the field has our standard input label, and also slightly changed the error spacing. you can see both changes in this before/after.

It used to look like this https://i.gyazo.com/a50266e8c77040a91123c931cca022cc.mp4
Now it looks like this https://i.gyazo.com/9b8a2a0ae033aec5e854c167cbaaeb49.mp4

src/components/RadioButtons/RadioButtons.js
```diff
@@ -98,64 +64,96 @@ RadioButton.propTypes = {
...
+  useEffect(() => {
+    const isSelectedValue = typeof selectedValue !== 'undefined'
+    if (onChange && isSelectedValue) {
+      onChange({ value: selectedValue, isAnswered })
+    }
+    if (formChangeHandler && isSelectedValue) {
+      // Ensure all validators get called
+      let errorMessage = validate(selectedValue)
+      errorMessage = errorMessage.length ? errorMessage : ''
+      setError(errorMessage)
```
kenning: Errors are only handled and display if this is inside of a form and has the `formChangeHandler` supplied. So if you supply a validator to a standalone radiobuttongroup it can't show an error.

src/components/RadioButtons/RadioButtons.js
roblevintennis: We have an `onSelect` for button select groups, and I believe to make these controlled inputs we need an `onChange`. Why wouldn’t we want to expose?

kenning: well... we don't have onchange or onselect for the other inputs right? 

i don't have an opinion whether inputs should or should not have public onchange handlers, but i think the inputs should be consistent and their props should be named consistently, hence why they would all go in the same folder, and why they should share the same state-handling logic for initialvalue, currentvalue, validationSuccess callback, the gamut of form-provided props, etc. that's way out of scope for this pr so i guess you could just change this to onSelect to assuage me. still seems weird though, and pretty clearly a result of ButtonSelectGroup having the word 'select' in its name.

src/components/RadioButtons/RadioButtons.js
roblevintennis: Fixed:

image

roblevintennis: >It's totally un-tabbable. Tabbing just skips over the field completely. 
Navigation is fixed and per the spec https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html and if you tab in and then use arrow keys, space/enter to select etc.

roblevintennis:
References

https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html
https://github.com/palmerhq/radio-group/blob/master/src/index.tsx#L92
https://www.w3.org/WAI/tutorials/forms/grouping/#radio-buttons

src/components/Inputs/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,293 @@
...
+  ...rest
+}) {
+  let initialSelected
+  if (currentValue || typeof currentValue === 'boolean') {
+    initialSelected = currentValue
+  } else if (initialValue || typeof initialValue === 'boolean') {
+    initialSelected = initialValue
+  }
+
+  const [selectedValue, setSelectedValue] = useState(initialSelected)
```
alexthecarp: I'm assuming initialSelected is ok to be `undefined` here ? The code flow seems to suggest it's possible to not hit any of the `if`s and it's not initialized to anything ?

src/components/Inputs/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,293 @@
...
+  const [isAnswered, setIsAnswered] = useState(false)
+  const resolvedValidator = validator ? validator : () => ''
+  const [getError, setError, , validate] = useErrorMessage(resolvedValidator)
+
+  const validationRadio = () => {
+    let errorMessage = validate(selectedValue)
+    errorMessage = errorMessage.length ? errorMessage : ''
+    setError(errorMessage)
+    if (formChangeHandler) {
+      // Update form with the new value and a falsy error message
```
alexthecarp: why falsy? can't errorMessage be set? what am I missing ?

src/components/Inputs/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,293 @@
...
+  // This key handling essentially allows us to comply with the spec. See their example:
+  // https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/js/radioGroup.js
+  // And Jared Palmer's React version which he obviously referenced w3's example:
+  // https://github.com/palmerhq/radio-group/blob/master/src/index.tsx#L92
+  const handleKeyDown = React.useCallback(
+    (event) => {
+      const last = allRadioValues.length - 1
+      const index = allRadioValues.findIndex((i) => i == selectedValue)
+      event.persist()
+      var flag = false
```
alexthecarp: nit: could we add a more descriptive name for flag? like 'stateChanged' or something .. not sure this makes sense, that's how I'm reading it.

src/components/Inputs/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,293 @@
...
+      function setPrevious() {
+        if (index === 0) {
```
alexthecarp: out of curiosity, as a JS noob, if we closed over `flag` and added `flag=true` here and in `setNext`, would this actually work?

src/components/Inputs/RadioButtons/RadioButtons.js
roblevintennis: Right, if none of the conditions hit it will be as if `useState()` or explicitly `useState(undefined)`

src/components/Inputs/RadioButtons/RadioButtons.js
roblevintennis: That's just a poor leftover comment. I'll remove it

src/components/Inputs/RadioButtons/RadioButtons.js
roblevintennis: Yeah, `keyActedOn`

src/components/Inputs/RadioButtons/RadioButtons.js
roblevintennis: Funny, this whole section was very much inspired by the referenced works, and your questions helped me to track down some things that needed change.

The `var` will always get "hoisted" to top of it's scope e.g. function in this case. This is fine here, but I've switched it to `let`. Now, if we were in a "block", `let` will work like you intuit from other languages (it'll be limited to the scope of that block, whereas `var` will just get hoisted and potentially confuse the hell out of all of us :) 

But yeah, I think since it's defined in outer most part of the function it sort of doesn't matter between `var` and `let`, as you mentioned, it'd be closed over and visible from any inner functions. That said, happy we were able to remove the `var` :)

alexthecarp: Lol, thanks for the thorough explanation! Appreciate it! I'm still somewhat confused, I don't want to limit the scope to the block, I want the same scope it currently has, but I wanted to be able to access it from the closure. In any case, not a big deal.

src/components/RadioButtons/RadioButtons.js
zach-ethos: I think we can even take out the `{...rest}`, unless we want it for some reason. I think it was only for redux-form.

src/components/Inputs/RadioButtons/RadioButtons.js
zach-ethos: @alexthecarp If you mean outside the arrow function, I think that would work too.

PR #149, opened 2019-11-08
Adds a prefilled BirthdateInput field example that is wrapped in a form.

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1148773948505985/f)
- Prefilled BirthdateInput with `initialValue` should enabled the form's Continue button

**Screenshots:**
image

image

### DIFF
--- a/src/components/Form/Form.md
+++ b/src/components/Form/Form.md
@@ -475,30 +475,6 @@ const questionGroups = {
   },
   two: {
     fields: {
-      questionTwo: {
-        component: (props, options) => {
-          return <TextInput {...props} />
-        },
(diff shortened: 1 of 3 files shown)


src/components/Form/Form.md
```diff
@@ -473,6 +474,23 @@ const questionGroups = {
     },
   },
   two: {
+    fields: {
+      questionFour: {
```
roblevintennis: I initially started this in position 4 (last question) and it worked, so I tried putting it here...still works, the button gets enabled.

PR #151, opened 2019-11-09
Deploy styleguide to eds.ethoslabs.io

kittyflat:
Currently accessible at http://eds.ethoslabs.io (will change to https)

TODO
- [ ] ~use https~ (separate PR)
- [x] bucket permissions for user/ci
- [x] ci config
- [x] figure out why `bin/deploy.sh` errors can still result in a pipeline success
- [ ] (after merge) update jenkins pipeline to use `master` branch instead of `deploy-styleguide`


ci/deploy.groovy
```diff
@@ -0,0 +1,20 @@
+pipeline {
+  agent { node { label 'Node' } }
+  tools { nodejs "10.16.3" }
+
+  stages {
+    stage('Run the deploy script') {
+      steps {
+        sh 'printenv'
+        sh 'bin/deploy.sh'
```
kittyflat: when testing the jenkins pipeline, i've noticed that if one of the commands fails in the `.sh` script, jenkins still shows the test as passed... i think any error in the `.sh` script should result in a pipeline failure. need to look into what's causing it to pass.

roblevintennis: yes that's odd is the nonzero status code getting gobbled up because these are somehow getting wrapped in an "outer script" or something?

roblevintennis:
Is this really ready for review @kittyflat ?

> figure out why bin/deploy.sh errors can still result in a pipeline success

Would you like to do that as a follow up on another PR or something?

kittyflat:
> Is this really ready for review @kittyflat ?

sorry, i thought it was ready but then i thought of these issues that i wanted to address in this PR

ci/deploy.groovy
heisian: might be a good idea to see what happens when `exit 1` is placed in the bash script.

kittyflat: `exit 1` in bash script properly exits the pipeline. if a command in the bash script exits with error code, we need to check for it and exit. i've added checks for important parts of the script.

bin/deploy.sh
```diff
@@ -0,0 +1,61 @@
...
+# Sync to S3, delete files that exist in destination but not in source,
+# except for fonts.zip.
+aws s3 sync ./styleguide s3://eds.ethoslabs.io/ --delete --exclude "fonts.zip"
+# Note: This check can cause the script to error when run by user without
+# aws delete permissions.
+if [[ $? -ne 0 ]]; then
+  echo "Error: \`aws s3 sync\` exited with error code."
+  echo "Cannot continue with deployment."
+  exit 1
+fi
```
kittyflat: no users (including admins) have delete permissions on the s3 bucket, so this check will make the script error when run locally. it will only succeed in CI

bin/deploy.sh
```diff
@@ -23,19 +23,29 @@ yarn install --check-files --production=false
 # Download fonts from S3 bucket
 aws s3 cp s3://eds.ethoslabs.io/fonts.zip bin/fonts.zip
 if [[ ! -f "bin/fonts.zip" ]]; then
-  echo "Error: bin/fonts.zip does not exist."
+  echo "Error: \`bin/fonts.zip\` does not exist."
```
roblevintennis: This is interesting to me. Does it come out looking "code formatted" to stdout?

bin/deploy.sh
```diff
@@ -23,19 +23,29 @@ yarn install --check-files --production=false
...
+  exit 1
+fi
 
 # Sync to S3, delete files that exist in destination but not in source,
 # except for fonts.zip.
 aws s3 sync ./styleguide s3://eds.ethoslabs.io/ --delete --exclude "fonts.zip"
+if [[ $? -ne 0 ]]; then
+  echo "Error: \`aws s3 sync\` exited with error code."
+  echo "Cannot continue with deployment."
+  exit 1
```
roblevintennis: So I take it that you had to explicitly exit nonzero because `$?` gets absorbed into the groovy wrapper correct?

bin/deploy.sh
kittyflat: no just comes out like this: \`wateverz\`

bin/deploy.sh
kittyflat: i think bc it's all running within this script, it is absorbed by this script, so this script needs to exit

bin/deploy.sh
roblevintennis: gotcha

zach-ethos:
I'm getting a 403, did something change?

kittyflat:
Yea, working on cloudfront updates atm

kittyflat:
It's been restored http://eds.ethoslabs.io

PR #153, opened 2019-11-11
Require a version bump on each PR

zach-ethos:
This aims to reduce versioning confusion downstream by requiring EDS PRs to increment the package.json version. This is to prevent multiple yarn.lock hashes for the same package version.

Here's there error you'd see on PR checks, before and after running `yarn release`:

Capture d’écran 2019-11-11 à 14 06 01
Capture d’écran 2019-11-11 à 14 08 52


roblevintennis:
I like this approach, I just thought that @kittyflat was going to convert this all to Jenkins in which case it might make sense for that to happen and all this integrate with Jenkins. Jason, can you provide some thoughts here?

PR #159, opened 2019-11-12
Update README.md

zach-ethos:


kenning:
thanks zach

package.json
```diff
@@ -1,18 +1,18 @@
...
     "css:dev": "chokidar src/**/*.scss -c 'yarn build' --initial",
+    "css:format": "prettier --write \"src/components/**/*.*css\"",
     "cy:open": "cypress open",
     "cy:run": "cypress run --browser chrome --config video=false",
     "e2e:ci": "start-server-and-test styleguide http://localhost:9008 cy:run",
     "e2e:dev": "start-server-and-test styleguide http://localhost:9008 cy:open",
-    "format": "prettier --write \"src/components/**/*.*css\"",
     "lint": "eslint src --max-warnings 0",
     "preversion": "yarn lint && yarn test && yarn build && yarn stats && git diff --exit-code",
     "spotcheck": "yarn build && echo \"\n*****\n* View the spotcheck at http://localhost:8080/spotcheck.html\n*****\n\" && webpack-dev-server ",
```
roblevintennis: We got rid of `release` that creates the tags somewhere else, huh? We probably should just remove `spotcheck` at some point (not related to this PR per se)

jameshawkinsjr: It also looks like `spotcheck` is hosted at `http://localhost:8081/` vs. `http://localhost:8080/spotcheck.html`

package.json
```diff
@@ -1,25 +1,24 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.10",
+  "version": "1.1.12",
```
jameshawkinsjr: Need to up this, since it's behind `master` now

package.json
roblevintennis: yeah, no one's running that script I don't believe haha

README.md
```diff
@@ -3,36 +3,28 @@
...
+1. See `package.json` for node `engines` – you need to have that version of node installed specifically, e.g. via `nvm`:
 
 ```shell
-nvm install v10.16.3 # then do `nvm list` to ensure version is available
+nvm install v10.16.3  # then do `nvm list` to ensure version is available
+nvm use               # use the version specified in `.nvmrc`, if installed
 ```
 
-## Installing fonts for local development
+2. Download `fonts.zip` (pinned to #dev-design-system) and unzip it.
```
roblevintennis: I think this is now in Google Drive since Slack deletes pinned stuff after a couple of weeks.

roblevintennis: https://drive.google.com/drive/u/0/folders/1hvAAEUWEsz2Hq-Wmj09xOvCYeixkJ4_c

README.md
```diff
@@ -69,13 +61,31 @@ The Cypress application will open up and it should be self evident how to run th
 
 _Our CI/CD is setup via Travis—you can find its configuration at `.travis.yml`_
 
-## Verify that the Design System works within CMS
+## Usage
+
+To use a release downstream, update your app's `package.json` as follows:
```
roblevintennis: I don't know if we should mention, but for monorepo, if one team sets EDS to `v1.2.3` and another team needs a `v1.2.4`, the first team will have to deal with the fact that EDS will get bumped. So, `v1.2.3` should not be considered as a lock for that project, but a convenient means to avoid keeping track of exact sha1's

zach-ethos: Not sure what you mean. Seems like CMS and monorepo could be on different versions.

README.md
zach-ethos: Fixed.

README.md
```diff
@@ -12,7 +12,7 @@ nvm install v10.16.3  # then do `nvm list` to ensure version is available
 nvm use               # use the version specified in `.nvmrc`, if installed
 ```
 
-2. Download `fonts.zip` (pinned to #dev-design-system) and unzip it.
+2. Download `fonts.zip` from [Google Drive](https://drive.google.com/drive/u/0/folders/1hvAAEUWEsz2Hq-Wmj09xOvCYeixkJ4_c) and unzip it.
```
roblevintennis: 👍

PR #160, opened 2019-11-13
Update UniversalNavbar to accept logoHref parameters

jameshawkinsjr:
**Description:**
- Because the `UniversalNavbar` is used in the monorepo (`app.ethoslife.com`) and the CMS (`ethoslife.com`), having the logo route to `/` sends the user to different pages, depending on whether you're in the main app or CMS.
- This allows for passing in a parameter, `logoHref` to the component, so that we can send a user directly to `https://ethoslife.com/` rather than `href='/'`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.15",
+  "version": "1.1.16",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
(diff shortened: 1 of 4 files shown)


jameshawkinsjr:
Alternatively, we can have this be a different route than `/index`, however we would then need to create a custom route in the main app and the CMS that routes to the homepage. I was thinking `/home` but that's already reserved in the CMS, so `/index` seemed the cleanest.

PR #162, opened 2019-11-19
React select port

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1146751046397447/f)
- EDS Select and MultiSelect

This is a porting of [React-Select](https://github.com/JedWatson/react-select) (but v3) 
 from `frontend/shared/fields/StyledReactSelect.js` in MonoRepo.

**Screenshots:**
image

With menu open:

image


src/components/Select/Select.module.scss
```diff
@@ -0,0 +1,102 @@
+/* Same as Type.scss Body. Should we use a mixin? */
```
zach-ethos: Could probably also use CSS variables

roblevintennis: I'll look into that tomorrow 👍 

Also, I need to work on this a bit more:

image

I removed margin-top but that won't work for above use case where we wrap (the screengrab has the 2px margin-top so I'll have to add that back and then work out another way to get things looking correct)

roblevintennis: Ok, 65e937bc55b5ed9e23c17794fe0c144b7a4434de should address all of these

src/components/Select/Select.js
```diff
@@ -0,0 +1,33 @@
...
+  // loadOptions should take an inputValue and return a Promise that resolves
+  // to an array of options.
+  loadOptions: PropTypes.func,
+  onChange: PropTypes.func,
+  isAsync: PropTypes.bool,
+  className: PropTypes.string,
+}
+
+Select.defaultProps = {
+  classNamePrefix: 'StyledReactSelect', // for styling
```
zach-ethos: 🤔

roblevintennis: Lol, I know, pre-existing in the port. I figured we could live with this prefix but agree 🤔

kenning:
looks pretty slick

src/components/Select/Select.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ReactSelect from 'react-select'
+import ReactSelectAsync from 'react-select/async'
+
+import styles from './Select.module.scss'
+
+export const Select = ({ className, isAsync, ...rest }) => {
+  const props = {
+    className: `${className ? className : ''} ${styles.root}`,
```
kenning: why do we need classname?

src/components/Select/Select.js
```diff
@@ -0,0 +1,33 @@
...
+    ...rest,
+  }
+  if (isAsync) {
```
kenning: i'd love some documentation on when to use isAsync. i don't even know what it does

kenning: oh is that what you use loadOptions for?

src/components/Select/Select.module.scss
```diff
@@ -0,0 +1,104 @@
+.root {
+  --remove: #d6d6d6;
+  --remove-hover: #c6c6c6;
```
kenning: very nitpicky but i would like a broader variable name for these as they are global.

src/components/Type/Type.module.scss
```diff
@@ -107,10 +107,9 @@
 }
 
 .Body {
-  $font-size: 19;
-  $line-height: 24;
-  font-size: $font-size * 1px;
-  line-height: $line-height / $font-size;
+  --base-line-height: 24;
+  font-size: calc(var(--base-font-size) * 1px);
+  line-height: calc(var(--base-line-height) / var(--base-font-size));
```
kenning: i believe i had issues mixing some combination of calc, scss variables, and css variables in some eds-> monorepo or cms deployment process at some point. i would recommend being careful about this, and possibly trying to jump through all the hoops to pull this into cms and make sure `yarn build` runs, as well as pulling this into monorepo and deploying it to a sandbox or staging server to avoid the crazy 'scss comments' blocker that we had about a month ago. 

alternatively you could just do `font-size: 19px; line-height: 1.2632;` but where's the fun in that 😛

roblevintennis: Those are defined in 

`src/components/Type/Type.scss` and the index: `src/components/index.scss` in turn imports that into the global css that results in design-system.css so I don't see this being an issue without all of our globals being an issue.

src/components/Select/Select.js
roblevintennis: At the top of the markdown file I do link to the library:

```md
We leverage [React-Select](https://github.com/JedWatson/react-select) for our single and multi selects. For example, you can add
`menuIsOpen={true}` below as an attribute to `<Select>` and the menu will remain open.
```

I'll add a link in the js file too though.

src/components/Select/Select.js
roblevintennis: I'm porting over previous version from monorepo and I'm not as worried about misuse of customizations, so I'll leave it.

PR #166, opened 2019-11-21
Possibly controversial readme update

kenning:
This is a rough draft and hits most of the concepts that seem to confuse people who use the EDS but haven't developed in it. 

I think it's important to get the messaging on this right so people can understand what the unwritten rules of the EDS design are. The first step is for us to all agree on them, this is a very rough draft and seems likely to create disagreements. My goal here was mostly to record things as they are now, without providing any "we'd like to do X eventually" commentary.


README.md
```diff
@@ -89,3 +89,25 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
 $ git push -d origin v1.2.3         # delete tag v1.2.3 remotely
 ```
+
+## Design patterns
+
+#### JS concepts
+
+- The repo is set up to import individual components, which then in turn import their own scss modules. 
+- Preference for hooks and functional components over classes.
+- Preference for proptypes validation over typescript and .d.ts files.
```
kenning: At this point I'd say not having .d.ts files is a bit unusual for a js library. I'm not trying to make a statement here, just saying that we prefer proptype validation.

kenning:
You can see a preview of the markdown here: https://github.com/getethos/ethos-design-system/tree/design-patterns-readme

README.md
```diff
@@ -89,3 +89,25 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
 $ git push origin v1.2.3            # push tag v1.2.3
 $ git push -d origin v1.2.3         # delete tag v1.2.3 remotely
 ```
+
+## Design patterns
+
+#### JS concepts
+
+- The repo is set up to import individual components, which then in turn import their own scss modules.
```
roblevintennis: > which then in turn import their own scss modules.

This seems potentially confusing if the reader is thinking about this in terms of a specific component. I'm fine with the part up to the comma

README.md
roblevintennis: I don't want to assert this as we may port EDS into typescript and I don't feel this assertion is necessarily helpful. If someone wants to actually setup tsconfig and a .d.ts file I'm for it.

README.md
```diff
@@ -89,3 +89,25 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
 ```
+
+## Design patterns
+
+#### JS concepts
+
+- The repo is set up to import individual components, which then in turn import their own scss modules. 
+- Preference for hooks and functional components over classes.
+- Preference for proptypes validation over typescript and .d.ts files. 
+- Please try to avoid using `<Media>` and `<Spacer>` components for now, they will be deprecated soon. For media queries, import `@include` rules from the media scss file as described below.
```
roblevintennis: > they will be deprecated 

replace with

> they are deprecated

README.md
```diff
@@ -89,3 +89,25 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
+
+#### CSS/SCSS concepts
+
+- CSS variables are the core of styling in the EDS. 
+  - All CSS variables are available on all pages in both monorepo and cms; they are the single source of truth for brand colors, breakpoints, and other commonly reused bits of css.
+- The EDS is (to an extent) designed as a set of immutable components in terms of interior JS logic as well as styling. New styling of existing components should be handled by PR's on the EDS, not by writing local overrides in files. Due to the generated hashed classnames from scss modules, it is almost impossible to sidestep styling rules of the EDS repo from monorepo or cms.
```
roblevintennis: I'm fine with this, but less the last sentence:

> Due to the generated hashed classnames from scss modules, it is almost impossible to sidestep styling rules of the EDS repo from monorepo or cms.

Let's take that out. It's implementation detail and I would also argue from experience that it is in fact possible to still override (although of course we don't want folks to do so)

README.md
```diff
@@ -89,3 +89,25 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
+- We encourage a limited subset of SCSS features for the EDS.
+  - The primary reason for using scss is for the [nesting syntax](https://sass-lang.com/guide#topic-3). 
+  - The secondary reason for using scss is to import @include statements from the media scss file for reusable media queries, like this:
```
roblevintennis: Not sure these are the primary and secondary reasons per se. @zach-ethos do you agree with these assertions?

In terms of the primary reason, I personally hate sass's ease of nesting and it causes all sorts of overly specific resulting CSS code.

Secondary reason seems to make sense.

I don't see the value of making the last statements regarding using `calc`, but we can leave it in here for now if you feel strongly. Ultimately, all of this stuff will get ported into some sort of Guidelines area of the EDS https://app.asana.com/0/1137984670709817/1137983928373132/f

README.md
```diff
@@ -89,3 +89,24 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
 $ git push -d origin v1.2.3         # delete tag v1.2.3 remotely
 ```
+
+## Design patterns
+
+#### JS concepts
+
+- The repo is set up to import individual components as necessary.
+- Preference for hooks and functional components over classes.
+- Please try to avoid using `<Media>` and `<Spacer>` components for now, they are deprecated soon. For media queries, import `@include` rules from the media scss file as described below.
```
zach-ethos: ~soon~

README.md
```diff
@@ -89,3 +89,24 @@ $ git tag -d v1.2.3                 # delete tag v1.2.3 locally
...
+#### CSS/SCSS concepts
+
+- CSS variables are the core of styling in the EDS. 
+  - All CSS variables are available on all pages in both monorepo and cms; they are the single source of truth for brand colors, breakpoints, and other commonly reused bits of css.
+- The EDS is (to an extent) designed as a set of immutable components in terms of interior JS logic as well as styling. New styling of existing components should be handled by PR's on the EDS, not by writing local overrides in files. 
+- We encourage a limited subset of SCSS features for the EDS.
+  - The primary reason for using scss is to import @include statements from the media scss file for reusable media queries, like this:
+    - `@import '~ethos-design-system/src/components/Media/Media.scss';`
+    - `@include for-phone-and-tablet {`
+  - Another accepted feature of scss is [nesting syntax](https://sass-lang.com/guide#topic-3).
```
zach-ethos: I'd strike this bullet too.

roblevintennis:
you'll need to bump the version I think

kenning:
Something screwed up when i bumped it to 1.1.20 and things got kind of weird. Mind if it goes to 1.1.21?

roblevintennis:
You should pull the very latest then bump version. I think it's up to .23 so you'll actually need to bump to .24:

https://github.com/getethos/ethos-design-system/blob/master/package.json#L3

PR #167, opened 2019-11-21
Deploy EDS with GH actions

kittyflat:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,59 @@
...
+      - master
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install dependencies
+        run: yarn install --production=false
```
kittyflat: can try caching this

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Download fonts from S3
+        uses: actions/aws/cli@master
```
kittyflat: building this action takes > 1m but not sure how to cache this

kittyflat:
still "room for improvement" with caching but it's also working now

.aws/config
```diff
@@ -0,0 +1,4 @@
+[profile ci-user]
+role_arn = arn:aws:iam::248295656887:role/ci-user
```
roblevintennis: Should we be exposing `248295656887` here? This is a public repo

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Download build files
+        uses: actions/download-artifact@master
+        with:
+          name: styleguide
+      - name: Sync styleguide to S3
+        uses: actions/aws/cli@master
+        with:
+          args: s3 sync ${GITHUB_WORKSPACE}/styleguide s3://eds.ethoslabs.io/ --delete --exclude "fonts.zip" --profile ci-user
```
roblevintennis: Does `sync` and `delete` here basically equate to an upsert where we're overwriting with new styleguide correct?

.aws/config
kittyflat: good point, let me see if there's some way to remove it from the repo

.github/workflows/deploy.yml
kittyflat: yeah `--delete` means delete everything that doesn't match the source, `--exclude` except for "fonts.zip"

PR #172, opened 2019-11-27
Add optional title to Select component

jameshawkinsjr:
**Description:**
- Adding an optional `title` parameter to `Select` that appears over the input box.
- ✅ Size of input box doesn't change when title is added

- 

**Screenshots:**
Screen Shot 2019-11-27 at 11 57 08 AM
Screen Shot 2019-11-27 at 11 57 01 AM
Screen Shot 2019-11-27 at 11 56 54 AM
Screen Shot 2019-11-27 at 11 56 45 AM
title2


src/components/Select/Select.module.scss
```diff
@@ -15,9 +15,7 @@
 
     .StyledReactSelect__multi-value {
       margin: 0;
-      padding: 3px;
-      margin: 0;
-      padding: 3px;
+      padding: 2px;
       margin-left: 0;
```
jameshawkinsjr: Only change I made to the existing styling:
1) You had duplicate styles in here
2) I made the padding for the multi-value boxes a tiny bit smaller, so it could fit without looking too crowded.

Let me know if that wont work and I can try to see if I can move around without changing this padding

PR #174, opened 2019-11-29
Adds Xhr and Nora sections and renames to EDS

roblevintennis:
**Description:**
- [Asana—EDS navigation structure](https://app.asana.com/0/1137984670709817/1151809971022172/f)
- Alphabetically orders the components in the left nav (it was horribly out of order before because styleguidist doesn't support alphabetical ordering once  you start using sub-directories)
- Core, **Nora**, Xhr, and potentially other supported platforms are now separated in the left navigation hierarchy:

tmp

This is very important as it will allow developers to distinguish between a component that's intended for use in the core product (e.g. insurance app) vs. one that might be just intended for use in Nora, or Marketing. Also, it will allow for modules that are not UI components, but cross cutting concerns. Xhr is a module for making http requests and handling server errors at the API request intersection gracefully. This really needs to be sharable to both _monorepo_ and _CMS_ (until CMS gets pulled into monorepo which is probably still quite a ways off). By locating such modules here in EDS, we will benefit as:

- CMS and monorepo can both leverage the Xhr component
- We start to establish a single source of  truth for shared and reusable front end components and modules


src/nora/components/Example/Example.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
```
roblevintennis: These are just intended to be placeholder components to be replaced by "real ones" once we start to actually use these

styleguide.config.js
```diff
@@ -33,6 +33,16 @@ module.exports = {
       components: './src/components/**/*.js',
       sectionDepth: 2,
     },
+    {
+      name: 'Xhr',
+      content: './src/xhr/content.md',
+      components: './src/xhr/components/**/*.js',
+    },
+    {
+      name: 'Nora',
```
roblevintennis: With this approach, we could potentially also add something like Marketing, perhaps for any off-brand components that are only meant for usage in landing pages and such.

src/components/ButtonSelectGroup/OptionButton.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import uuidv4 from 'uuid/v4'
-import { Button } from '../../Button'
+import { Button } from '../Button'
```
roblevintennis: Note: Having things like `Inputs/ButtonSelectGroup` confused react-styleguidist in terms of it's ability to alphabetize the navigation. With more effort, we could override some modules in react-styleguidist, but I've elected to just push up things like `Inputs/ButtonSelectGroup`  to top level: `ButtonSelectGroup` and so much of this PR is just accommodating for these changes in paths.

tkarssli:
All nice changes!

PR #175, opened 2019-11-30
Xhr

roblevintennis:
# Should go in after https://github.com/getethos/ethos-design-system/pull/174 and base branch is pointed back to `master`

**Description:**
- [Port Xhr to EDS](https://app.asana.com/0/1150980499859961/1150980499859976/f)
- [Convert PathBuilder to Interface](https://app.asana.com/0/1150980499859961/1150982726068609/f)

_Ports over the FYSTS to EDS and sets up Typescript compilation local to the src/xhr directory._

- https://github.com/getethos/ethos-design-system/pull/174

**Screenshots:**

GET / DELETE / POST examples page:

image

image

Error handling examples:

image

image

image

PUT example:

image


src/xhr/components/PostGetDeleteExample.md
```diff
@@ -1,6 +1,15 @@
-POST | GET | DELETE Examples
+Click the `Get Posts` button below to trigger a GET request, then hover an item and
+click its 'X' to trigger a DELETE a particular item. You can also use the
+_Create a Post_ form at the bottom to trigger a POST and create a new item:
 
-```jsx
+```jsx noeditor
```
roblevintennis: This example snippet will no longer be shown, but will result in the component rendering. The code fence with `js { "fille": ...` will, however, use the entire file so folks can see how we used the _Xhr_ components and hooks:

image

styleguide.config.js
```diff
@@ -22,6 +23,28 @@ module.exports = {
   getExampleFilename(componentPath) {
     return componentPath.replace(/\.js$/, '.md')
   },
+
+  // This callback allows us to display the source code of a file within markdown:
+  // ```js { "file": "../mySourceCode.js" }
+  // Reference:
+  // https://react-styleguidist.js.org/docs/cookbook.html#how-to-display-the-source-code-of-any-file
+  updateExample(props, exampleFilePath) {
```
roblevintennis: Allows https://github.com/getethos/ethos-design-system/pull/175/commits/56a35489bfb82e31cdd605569ae45688715a30d6#diff-d94af5a1dbc45e3bc43a4c03b164774dR14 to work in pulling in the entire file as a code example

src/xhr/lib/errorReporter.js
```diff
@@ -0,0 +1,34 @@
...
+// TODO: Add in Sentry
+export default {
+    captureException: (errorOrResponse) => {
+        if (process.env.NODE_ENV === 'production')
+            return;
+        let name = 'XhrError';
+        let message;
+        let response;
+        let stack = [];
+        if (errorOrResponse instanceof XhrError) {
```
roblevintennis: This is a slight departure from where we left FYSTS upon merging https://github.com/getethos/FYSTS/pull/7 ... it turned out checking for `typeof errorOrResponse.ok` against `errorOrResponse` when it's an instance of `XhrError` was angering Typescript in this environment since `XhrError` has no `.ok` property. Changing it to this made more sense to me anyway, and removed the compilation error as well so I went with it.

src/xhr/lib/xhr.js
```diff
@@ -0,0 +1,98 @@
+import errorReporter from './errorReporter';
```
ekeric13: this code is tough to follow

src/xhr/lib/xhr.js
```diff
@@ -0,0 +1,98 @@
...
+        const resolvedPath = typeof path === 'string' ? path : path.buildPath();
+        // Create a new XhrRequest object that will be passed into browser fetch
+        const request = new XhrRequest(`${baseURL}/${resolvedPath}`, xhrOptions);
+        // Begin timing how long our request will take
+        request.setTimestamp();
+        // Log request
+        xhrLog.request(request);
+        fetch(request)
+            .then((rawResponse) => {
+            return parseResponseBody(rawResponse, xhrOptions);
```
ekeric13: we seem to parse the body and then see if it is "ok". I think it makes sense to check if the response was successful to begin with

src/xhr/lib/xhr.js
```diff
@@ -0,0 +1,98 @@
...
+        })
+            // Use the XhrLog to log everything going on in the Xhr layer
+            .then((xhrResponse) => {
+            xhrLog.response.bind(xhrLog)(xhrResponse, request.timestamp);
```
ekeric13: not sure why this needs it's own "then" block

src/xhr/lib/xhr.js
```diff
@@ -0,0 +1,98 @@
...
+            else {
+                // Set err and report if request was unsuccessful
+                err = new XhrError(xhrResponse);
+                errorReporter.captureException(xhrResponse);
+            }
+            resolve({ err, response });
+        })
+            // Handle network errors
+            .catch((xhrError) => {
+            // Log the error with our reporter
```
ekeric13: do we ever get here? We seem to never throw

src/xhr/lib/xhr.js
roblevintennis: I dunno, this was Tim's code and it seems he's taking the raw response, converting it to `xhrResponse` which in TS land is backed by an interface. Then, he checks the `.ok` status. We might be able to simplify some of this in a later iteration I agree.

roblevintennis: Also, I think it's more useful to look at the .ts code as this is generated

PR #176, opened 2019-12-01
Svg sprite loader

roblevintennis:
# Note: We'll want to update all consumers that import <Logo /> -- with corresponding Webpack / Svg-Sprite-Loader setup

I want to stay away from bundling things, so this svg-sprite setup is only really for `--dev` so I think although it works here, monorepo pulling in the LogoInline will break. So, I think we need to also do this in monorepo (setup svg-sprite-loader), and then point to this branch, verify the LogoInline uses are rendering properlly, and then merge this, and then repoint monorepo back to `master`.

https://github.com/getethos/ethos/pull/1721

**Description:**
- [Asana Task](https://app.asana.com/0/1136962714401929/1137983928373125/f)

- Brings in [svg-sprite-loader](https://github.com/JetBrains/svg-sprite-loader/blob/HEAD/2.0.md) so we can start to use svg sprites where appropriate as opposed to always doing inline svg.
- Only converts the inline logo to use SVG sprite so far.

_For those unfamiliar with this approach, I believe the screenshot below should make it more evident how it works_

- [x] Fancy logo on UniversalNavbar in mobile when scrolled down still animates and svgs work
_Note, I don't see why we shouldn't be able to iterate and convert those individual letter SVGs to use inline too maybe_

- [x] ValueProps svgs render:

image

- [x] Button arrows still render (also should be converted in a later iteration):

image

- [x] Cloudinary images SVG examples still renders (it  looks poor but same off master https://eds.ethoslabs.io/#/Components/CloudinaryImage):
image

 
**Screenshots:**

Here's `LogoInline` (only SVG converted so far) using the sprite:

svg-sprite-example


webpack.config.js
```diff
@@ -63,18 +64,47 @@ module.exports = {
...
       {
-        test: /\.svg$/,
-        loader: 'file-loader',
-        query: {
-          name: 'static/media/[name].[hash:8].[ext]',
-        },
-      },
-      {
-        test: /\.(woff|woff2|eot)$/,
+        test: /\.(woff|woff2|eot|png|jpe?g|gif)$/,
```
roblevintennis: Seems like we should also just support gif, png, jpg/jpeg

webpack.config.js
```diff
@@ -63,18 +64,47 @@ module.exports = {
...
+                { removeTitle: true },
+                { removeNonInheritableGroupAttrs: true },
+                { removeUselessStrokeAndFill: false },
+                { collapseGroups: true },
+                { convertColors: { shorthex: false } },
+                { convertPathData: false },
+                {
+                  removeAttrs: {
+                    attrs: '(fill|stroke)',
+                    preserveCurrentColor: true,
```
roblevintennis: this options allows you do something like `<path fill="currentColor"...` and svgo will preserve or not remove those kinds of attributes. Otherwise, the whole point of SVGO is to remove as many attributes uneeded as possible to optimize

webpack.config.js
```diff
@@ -63,18 +64,47 @@ module.exports = {
...
+          {
+            loader: 'svg-sprite-loader',
+          },
+          {
+            loader: 'svgo-loader',
+            options: {
+              plugins: [
+                { removeTitle: true },
+                { removeNonInheritableGroupAttrs: true },
+                { removeUselessStrokeAndFill: false },
```
roblevintennis: The idea here is you opt in for removal of any presentational attributes with the idea that you'll use external CSS styles to account for these

webpack.config.js
```diff
@@ -63,18 +64,47 @@ module.exports = {
...
+              ],
+            },
+          },
+        ],
+      },
+      // Any non-sprited SVG images or similar use file-loader
+      {
+        test: /\.svg$/,
+        exclude: /node_modules/,
+        include: [],
```
roblevintennis: TODO: I believe we'll need to "add back" any SVGs we do NOT want pulled in via SVG sprites e.g. anything we want to leave inline or pulled in via css background images etc. etc.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.24",
+  "version": "1.1.27",
```
roblevintennis: There are 25 and 26 PRs which I anticipate going in before this

src/svgs/logo-inline.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 971.997 180.498">
+  <title>Ethos Logo</title>
```
jameshawkinsjr: It looks like we lose the title in this branch:

**Current:**
Screen Shot 2019-12-02 at 12 43 16 PM

**This branch**
Screen Shot 2019-12-02 at 12 43 31 PM

webpack.config.js
jameshawkinsjr: Maybe we can add documentation on the styleguide that includes what attributes are available via CSS (didn't see that `removeTitle` was checked as "True" -- wouldn't we want to default to include the title for accessibility purposes?

roblevintennis: Ahh yes, so removing the title is something that folks do for large sets of icons that are  meant to be decorative (https://dev.to/adammcquiff/how-and-why-to-clean-svg-markup-49i shows a bit about that), but, since we're sort of straddling the lines likely, I'll remove this rule so it'll default to false. If we leave it as false, and start to add a bunch of icons, we'll either want to remove titles later, or, ensure we at least provide useful titles as they'll get announced by screen readers

roblevintennis: But good call! So, I'm about to push up a commit now that leaves <title> tags intact:

image

PR #177, opened 2019-12-02
Add ValueProps to index.js

tkarssli:
Add ValueProps to index.js so we can use it


src/components/index.js
```diff
@@ -6,6 +6,7 @@ export { Layout } from './Layout'
 export { Media } from './Media'
 export { Select } from './Select'
 export { Spacer } from './Spacer'
+export { ValueProps } from './ValueProps'
```
roblevintennis: 👍 try running `yarn version` and it'll guide you to bump the version number...without looking I'm pretty sure that's the CI/CD issue

PR #182, opened 2019-12-05
Adds react-select/creatable and react-select/async-creatable to Select Component

wafflehorse:
**Description:**
- Will be used in this ticket: [Special Auth in flow](https://app.asana.com/0/1149628124018183/1150703512578001)
- Adds react-select/creatable component by providing isCreatable prop to our Select component.
- Also adds react-select/async-creatable for the case where both isAsync and isCreatable are passed as props.

**Screenshots:**
Screen Shot 2019-12-04 at 6 20 54 PM

### DIFF
--- a/src/components/Select/Select.md
+++ b/src/components/Select/Select.md
@@ -1,9 +1,10 @@
 We leverage [React-Select](https://github.com/JedWatson/react-select) for our single and multi selects. For example, you can add
 `menuIsOpen={true}` below as an attribute to `<Select>` and the menu will remain open.
+
 ```jsx
-const onSelected = selectedOption => {
+const onSelected = (selectedOption) => {
   console.log('Option selected: ', selectedOption)
-};
+}
 const options = [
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.md
```diff
@@ -1,9 +1,10 @@
 We leverage [React-Select](https://github.com/JedWatson/react-select) for our single and multi selects. For example, you can add
```
wafflehorse: is this what we expect our prettier config to do? Or is my setup messed up?

roblevintennis: It's a bit odd, we have `"singleQuote": true,` so I'd think it'd be single quotes

wafflehorse:
@roblevintennis Thanks for the quick review! I bumped the version and push a new git tag. What does the git tag do?

roblevintennis:
@wafflehorse consumers point to the git tag

PR #183, opened 2019-12-05
Columns v0

kenning:
## Preface: Whoopsie

Due to what i've been told was a clerical error, i've lost access to gmail, slack and ethos github access ahead of my scheduled end date of this Friday. Drew said he was looking into getting some of them back online but luckily this repo is public so i can make a PR anyway. (Also luckily, I still have access to figma for some reason.) 

This also means that the (small) changes I made to the CMS are not able to be pushed if I haven't already (I forgot what I pushed on top of Drew's video PR and don't have access to it anymore) and I certainly can't integrate my changes here into the CMS and push anything up there.

## Description 

This idea for a new EDS component came about when I was adding video support for the Highlight module while visiting last week. I realized I was reimplementing layout CSS which felt suspiciously similar to the CSS for the quote widget, as well as some other modules i've seen in my limited time working on the CMS. I asked Ryan about this and we discussed recurring patterns of two columns with optional 64/80px gutters (laptop/desktop viewports) to the left/right of a column's content. Suffice it to say there are recurring patterns which have been manually re-implemented in many cms modules and even a couple things in the monorepo. 

The reason for this layout being introduced now and not way back when the EDS was originally started is actually that these designs 'evolved over time' (quote from Ryan here) and seems not to be written down anywhere, but is still a repeating pattern. I was actually against adding additional Layout elements at the beginning because I couldn't see consistent patterns; I'm not sure if that was a mistake or not, but there's definitely consistent patterns now.

Looking around at various CMS modules it becomes clear that, despite (or more likely, due to) spending a ton of time doing similar-but-not-exactly-the-same layouts, we actually have a fair number of inconsistencies and little design issues that are live right now. You can see these for yourself by starting up the cms and navigating to /demo, placing your cursor exactly in the middle of a module that's supposed to cut the page evenly in half, and scrolling up or down until you find a module that also is supposed to cut the page in half but is off by ~5-10px. Here's some videos of that happening: https://i.gyazo.com/e7b38b50f9e1191e5b8a3d6b22b189b0.mp4 https://i.gyazo.com/ed56b6253e988cda41355063fdd68610.mp4
Highlight and Header Versatile modules seem to be the culprits here; Highlight was the module I was supposed to be working on originally, which is another reason I did this instead of just fixing that one module. 

The benefits of making a reusable component is not only to prevent repetitive implementation of the same or similar sets of code, but also to prevent subtle errors like this going forward. Layout.HorizontallyPaddedContainer is a great example of doing a layout component once and never having to worry about it again; ideally this would be similar.

## Instructions

Run `yarn styleguide` and go to the new 'columns' page to see four demos of loose reproductions of CMS modules. To see them you have to click the hot pink "show/hide button" in the upper right. I think the code and demos basically speak for themselves after that.

## Following up on this

I doubt that the code I've written here will get reviewed thoroughly before I finish at ethos on Friday. I also seriously doubt this implementation is where iteration on a module like 'columns' would end. I figured the most effective way to demonstrate the use of a 'columns' demo would be to reproduce the CMS modules that I think this will simplify, and during implementation I ended up adding features as I went (for example, the divider in between columns for the "Story" module). I imagine that we would want to replace our existing CSS for modules and add the occasional new feature to adapt to usage of this component when (if) it gets picked up, and asked Cole if he would be interested in taking this on should he get some spare time.

I'm not expecting this to be added or integrated into the CMS immediately, and with a new PM starting for growth team soon (and multiple 'pods' being shifted around and reassigned) I have no idea what the prioritization would be. 

Whenever someone picks this up I think it would be best to work very, very closely with design to make sure that what's happening is correct at every viewport. I've been cut off and it seemed not worth the time to get in contact with them before pushing up this early prototype, but the designers at Ethos always know exactly what they want to happen at every viewport even if the figma docs aren't 100% clear, and things work out great as long as they are included in the discussions, especially over video call or in person.

### DIFF
--- a/src/components/Colors.js
+++ b/src/components/Colors.js
@@ -12,5 +12,10 @@ export const COLORS = {
   GRAY_STROKE_AND_DISABLED: 'GrayStrokeAndDisabled',
   WHITE: 'White',
 
+  // Background colors
+  BRAND_DUCKEGG: 'BrandDuckegg',
+  BRAND_MOSS: 'BrandMoss',
+  BRAND_BUTTERCUP: 'BrandButtercup',
+
   // Miscellany
 }
--- a/src/components/Layout/Columns/ColumnWrapper.js
+++ b/src/components/Layout/Columns/ColumnWrapper.js
@@ -1,0 +1,8 @@
+import React from 'react'
+// import PropTypes from 'prop-types'
+
+import styles from './Columns.module.scss'
+
+export function ColumnWrapper({ children }) {
+  return <div className={styles.twoColumnLayout}>{children}</div>
+}
--- a/src/components/Layout/Columns/ColumnWrapper.js
+++ b/src/components/Layout/Columns/ColumnWrapper.js
@@ -4,6 +4,10 @@ import React from 'react'
 import styles from './Columns.module.scss'
 
 export default function ColumnWrapper({ children, divider }) {
+  if (React.Children.count(children) !== 2) {
+    throw new Error('ColumnWrapper can only be used with exactly two columns.')
+  }
+
   const classNames = [styles.twoColumnLayout]
   if (divider) {
     classNames.push(styles.dividerLine)
--- a/src/components/Layout/Columns/ColumnWrapper.js
+++ b/src/components/Layout/Columns/ColumnWrapper.js
@@ -3,6 +3,10 @@ import React from 'react'
 
 import styles from './Columns.module.scss'
 
-export function ColumnWrapper({ children }) {
-  return <div className={styles.twoColumnLayout}>{children}</div>
+export default function ColumnWrapper({ children, divider }) {
+  const classNames = [styles.twoColumnLayout]
+  if (divider) {
+    classNames.push(styles.dividerLine)
+  }
+  return <div className={classNames.join(' ')}>{children}</div>
 }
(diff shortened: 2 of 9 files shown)


src/components/Colors.js
```diff
@@ -12,5 +12,10 @@ export const COLORS = {
   GRAY_STROKE_AND_DISABLED: 'GrayStrokeAndDisabled',
   WHITE: 'White',
 
+  // Background colors
+  BRAND_DUCKEGG: 'BrandDuckegg',
+  BRAND_MOSS: 'BrandMoss',
+  BRAND_BUTTERCUP: 'BrandButtercup',
```
kenning: These came up at some point while I was mocking out cms modules, they aren't used now but there's no reason for them not to be available as variables (i think)

c7hudson:
If we're calling this version ready to implement we'll just need to run `yarn release` and add props validation to pass Travis CI, which is currently failing.

I can tackle those items later today if nobody else wants to take them on.

roblevintennis:
@c7hudson You probably weren't affected, but last week we had an issue with deployments due to a nuanced EDS dependency that was added. This resulted in backing down to 1.1.21. I'm re-introducing things and attempting to get things back to where we were. This needs to happen before we introduce anything new so I don't see this happening immediately.

Also, I think it's sometimes better to iterate and do follow ups so the PRs stay smaller. So for example, whatever it is your working on that's related, we could try to get that code complete, merged, and then, do a follow up to that once this PR is merged.

c7hudson:
All good, this feature can be written in after the work on my end.

kenning:
best of luck yall! glad to see it at least seems like a good direction 😄

PR #184, opened 2019-12-05
Revert svg stuff

roblevintennis:
No Asana. This is to deal with the issues with deployment over last couple of days. It appears as though the svgo svg-sprite-loader related stuff caused our deploy process to fail with a prompt for installing webpack-cli that looked something like:

```shell
$ /home/node/workspace/deploy/clients/node_modules/.bin/webpack --config ./src/admin/webpack.stage.js

One CLI for webpack must be installed. These are recommended choices, delivered as separate packages:

 - webpack-cli (https://github.com/webpack/webpack-cli)

   The original webpack full-featured CLI.

We will use "npm" to install the CLI via "npm install -D".

Do you want to install 'webpack-cli' (yes/no): Done in 1.46s.

...
```

- This PR reverts these two SVG related PRs:

https://github.com/getethos/ethos-design-system/commit/c3cb5b88d87aebac3218613145530051e73e6f37

https://github.com/getethos/ethos-design-system/commit/0447f8580c77e8b131af571a62a6c7b2524211e4


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.34",
+  "version": "1.1.35",
```
roblevintennis: Atm, I have to bump the version tag to get this to pass CI/CD (even though intuitively it feels  like I should be going back to a smaller number like 1.1.31 since I'm reverting stuff.

TODO: I think we should back out `ci/travis.spec.js` and rethink how we're doing all this cc / @zach-ethos @alexthecarp

roblevintennis: Yeah, I removed that spec to try to get EDS back to a known stable version which I believe is 1.1.31

PR #186, opened 2019-12-10
Fix styleguide:build after styleguidist upgrade

zach-ethos:


styleguide.config.js
```diff
@@ -70,5 +70,6 @@ module.exports = {
       sectionDepth: 2,
     },
   ],
+  assetsDir: ['src/assets'],
```
roblevintennis: lol, whatevs styleguidist

PR #190, opened 2019-12-13
Typescript defs

roblevintennis:
# After merging this, remember in monorepo to point back to a VERSION_NUMBER e.g #v1.1.99

**Description:**
- [Asana Task](https://app.asana.com/0/1152334753103228/1152842735737402/f)
- EDS Needs Better Typescript Support

### Note—much of this could stand to use more descriptive typing. In the interest of iterating and dealing with such a large set of changes, I would like to let these loose typings slide and then I will continue to make them stricter over time.

Monorepo PR: https://github.com/getethos/ethos/pull/1854

### DIFF
--- a/src/validators/BirthdateInputValidator.d.ts
+++ b/src/validators/BirthdateInputValidator.d.ts
@@ -13,8 +13,12 @@ export declare const dateStringMatchesFormat: (
   cleansedDateString: string,
   dateFormat: string
 ) => '' | 'Please enter a valid date.'
-export declare const getMaxDateValidator: (props: any) => (value: any) => any
-export declare const getMinDateValidator: (props: any) => (value: any) => any
+export declare const getMaxDateValidator: (
+  props: any
+) => (value: string | null | undefined) => string
+export declare const getMinDateValidator: (
+  props: any
+) => (value: string | null | undefined) => string
 export declare const validateMinMaxDateFactory: (
   props: any
 ) => (value: any) => any
--- a/src/validators/BirthdateInputValidator.d.ts
+++ b/src/validators/BirthdateInputValidator.d.ts
@@ -1,15 +1,20 @@
-export declare const DATE_FORMATS: string[];
+export declare const DATE_FORMATS: string[]
 export declare const dateMaskByFormat: {
-    'mm/dd/yyyy': (string | RegExp)[];
-    'mm/yyyy': (string | RegExp)[];
-    'mm/yy': (string | RegExp)[];
-};
+  'mm/dd/yyyy': (string | RegExp)[]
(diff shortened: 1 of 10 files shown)


src/validators/BirthdateInputValidator.d.ts
```diff
@@ -0,0 +1,20 @@
...
+export declare const dateRegexByFormat: {
+  'mm/dd/yyyy': RegExp
+  'mm/yyyy': RegExp
+  'mm/yy': RegExp
+}
+export declare const dateStringMatchesFormat: (
+  cleansedDateString: string,
+  dateFormat: string
+) => '' | 'Please enter a valid date.'
+export declare const getMaxDateValidator: (props: any) => (value: any) => any
```
alexthecarp: So I've been poking around, is there a better type we could use for validators? Would something like `(value: string): bool` or some such

src/validators/validateMinMax.d.ts
```diff
@@ -0,0 +1,6 @@
+declare const validateMinMaxFactory: (
+  min?: number,
+  max?: number,
+  message?: string
+) => (value: any) => string
```
alexthecarp: hmm so I'm not sure, this factory returns function that takes `any` and returns `string` ? why is the input just `number` ?

src/validators/BirthdateInputValidator.d.ts
roblevintennis: I think probably the input as string would be fine but I think it also returns a string (empty is valid, otherwise it contains the error message) so likely it would  be `(value: string) => string` if we wanted to change this.

src/validators/validateMinMax.d.ts
roblevintennis: I think Nick wrote this originally maybe? But `value` could  probably be a raw string or date and then it gets coerced into a string in the implementation. I don't want to change that in this PR though. `String(value` is what it does.

However, I should probably change the return to `string | undefined` to reflect `message : undefined` in  implementation.

roblevintennis: I think Nick wrote this originally maybe? But `value` could  probably be a raw string or date and then it gets coerced into a string in the implementation. I don't want to change that in this PR though. `String(value` is what it does.

However, I should probably change the return to `string | undefined` to reflect `message : undefined` in  implementation.

roblevintennis: Update, I've added 
4f98030 and bc6bd08 to make it a bit stronger typed, but I'm certain there are several places this sort of improvement could be made and I'd prefer to have something in and working and then iterate if possible. Does this look good enough?

PR #192, opened 2019-12-13
add faq to index.d.ts

tkarssli:
[Rob] Background: I had asked Tamir if he could add typescript typings for Faq and it resuled in this PR. I've piggy-backed off his work to make verification within EDS easier.

Faq in EDS:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "test:coverage": "jest --collectCoverageFrom '[\"src/components/**/*.js\", \"src/hooks/**/*.js\"]' --coverage && open ./coverage/lcov-report/index.html",
     "test:debug": "node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand",
     "test:update": "jest --updateSnapshot",
-    "test:scss": "./scss-test.sh"
+    "test:scss": "./scss-test.sh",
+    "test:types": "tsc --jsx react --noEmit -p ./src/components"
   },
   "jest": {
     "verbose": true,
@@ -51,6 +52,7 @@
     "@babel/plugin-proposal-class-properties": "^7.5.5",
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
+    "@types/react": "^16.9.16",
     "babel-eslint": "^10.0.3",
     "babel-jest": "^24.9.0",
     "babel-loader": "^8.0.6",
@@ -82,6 +84,7 @@
     "sass-loader": "^8.0.0",
     "start-server-and-test": "^1.10.5",
     "style-loader": "^1.0.0",
+    "typescript": "^3.7.3",
     "url-loader": "^2.1.0",
     "webpack": "^4.41.2"
   },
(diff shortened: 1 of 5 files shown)


roblevintennis:
This looks good. What I would suggest to determine if this is working or not is to attempt to consume it. So

1. Point `ethos-design-system` to this branch in monorepo
2. pull in the change
3. Put an import in one of the quote quiz files for the `faq` stuff
4. start new quote quiz following direction in frontend/src/quote-quiz/README.md
And just ensure it doesn't error

package.json
```diff
@@ -26,7 +26,8 @@
     "test:coverage": "jest --collectCoverageFrom '[\"src/components/**/*.js\", \"src/hooks/**/*.js\"]' --coverage && open ./coverage/lcov-report/index.html",
     "test:debug": "node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand",
     "test:update": "jest --updateSnapshot",
-    "test:scss": "./scss-test.sh"
+    "test:scss": "./scss-test.sh",
+    "test:types": "tsc --jsx react --noEmit -p ./src/components"
```
roblevintennis: This allows us to do `yarn test:types` and if no typescript errors, we've verified our typescript defs are working properly."

src/components/index.d.ts
```diff
@@ -2,7 +2,6 @@
 
 import React from 'react'
 declare class UniversalNavbar extends React.Component {
-  private props
```
roblevintennis: I found some issues unrelated to this PR, but, since I elected to use this opportunity to create an easier way to verify our typings, I think it _sort of_ IS related 😋

src/components/index.d.ts
```diff
@@ -74,8 +73,6 @@ export declare const Logo: {
   Inline: typeof LogoInline
 }
 
-import { Button } from './Button'
```
roblevintennis: This isn't needed and was also throwing an error

src/components/index.d.ts
```diff
@@ -611,6 +608,19 @@ export declare const NumberInput: {
   }
 }
 
+export declare const Faq: {
+  ({ questions, open }: { questions: any; open: any }): JSX.Element
```
roblevintennis: this used to be typed with `: boolean` but that was throwing an error `"JSX element type 'boolean' is not a constructor function for JSX elements."` and I noticed I've been having success with react component exports being of type `JSX.Element` which I believe is correct.

src/components/index.tests.tsx
```diff
@@ -0,0 +1,20 @@
...
+  {
+    summary: 'Second',
+    content: 'dos',
+  },
+]
+
+// Localized test
+class MyTest extends React.Component<any, any> {
+  render() {
+    return <Faq questions={demoQuestions} />
```
roblevintennis: And I suppose later we'll just wrap in a fragment like:

```tsx
render() {
  return (
    <>
       <TestFoo />
       <TestBar />
       { /* etc etc */ }
    </>
  )
}

roblevintennis: Update: I DID test it out also from the monorepo quotequiz (manually). See notes on  corresponding [PR](https://github.com/getethos/ethos/pull/1863)

tkarssli:
Signing off on additions

PR #193, opened 2019-12-17
[WIP - IDEA] Test for Index.js to verify that newly added files are not unintentionally left out of index.js

tkarssli:
Was just testing out this idea, if someone adds a new folder with a new component and follows current naming convention then tests will fail if they forget to add the component to export from `index.js` in `src`. If there is a added js file that should not be index then they can just update snapshot and push that along.

This is just and idea so if you think its not worth it or just plain wack I am fine with canning it. Just wanted some thoughts on it. Main purpose was to add a check for when we add new components and I forget to add to index and have to create a whole new PR to add it which could potentially delay deploys/ cause problems.


bin/verifyExports.js
```diff
@@ -0,0 +1,37 @@
...
+      file.includes('spec') ||
+      file.includes('test') ||
+      file.includes('index')
+    )
+  })
+
+  const componentsFromFiles = jsFiles.map((file) =>
+    file
+      .split('/')
+      .slice(-1)[0]
```
roblevintennis: what's `.slice(-1)[0]` doing here exactly?

roblevintennis: Also, if you're just trying to get `FooComponent` from `path/to/FooComponent.js` you could do
`path.basename('path/to/FooComponent.js', '.js')`

See https://nodejs.org/api/path.html#path_path_basename_path_ext

I'd prefer that over tricky splitting and array indicing

bin/__snapshots__/verifyExports.test.js.snap
```diff
@@ -0,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Index exports,   if a EDS component was recently added or removed, fix index then update snapshots 1`] = `
+Array [
+  "optionbutton",
```
roblevintennis: shouldn't there be more components in here?

bin/verifyExports.js
```diff
@@ -0,0 +1,37 @@
...
+      .split('.')[0]
+      .toLowerCase()
```
roblevintennis: Should we lower case the components? Not sure if we'll ever have a `foobar` and a `FooBar` but this would make them be considered as the same thing

tkarssli: Ah yeah and not only that we should be checking that exports from index are capitalized anyway. I almost accidentally exported a `faq` instead of a `Faq`. We could check this here

bin/verifyExports.js
tkarssli: ah Yeah this is the business. thank you

bin/__snapshots__/verifyExports.test.js.snap
tkarssli: Not sure about my approach here but this is basically leftover things, exports from index that dont actually have a .js file associated with it. But I see names in here that DO have .js files so something is broken.

PR #194, opened 2019-12-17
Accessible data grid

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1148603508892214/1146751046397446/f)
- EDS: Tables (aka Data Grids)...these will be used for __Nora__

This is the initial basic table implementation. But, it does already feature:
- A11y compliance with the [wai-aria-practices-1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#grid) specifications for data grids.
- keyboard navigation and roles are assigned to spec
- Boilerplate default styles
- Sort by column
- Responsive stacking at mobile and lower with column label on left and values on the right
- Adjustable row sizes (default aka medium), small, large

image


src/components/Grid/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Juan Carlos Medina
```
roblevintennis: Even though I practically rewrote this, I felt it was correct thing to leave his original license. See my thoughts here: https://github.com/getethos/ethos-design-system/pull/194/files#diff-515fd4503108cc3de9a70f41c8aba2dcR2

src/components/Grid/Column.js
```diff
@@ -0,0 +1,77 @@
+import PropTypes from 'prop-types'
+import React, { useEffect } from 'react'
+
+import RefType from './ref.js'
```
ekeric13: why are we importing it from ref.js and not from ref?

src/components/Grid/Grid.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types'
+import React, { Children } from 'react'
+
+import { Position } from './Position.js'
+import RefType from './ref.js'
+import styles from './Grid.module.css'
+
+export const Grid = (props) => {
+  const gridRefs = props.columnRefs.concat(props.rowRefs)
```
ekeric13: might destructure

roblevintennis:
Thanks @ekeric13 ... I've implemented changes per your two comments:

```shell
commit 89ad217786c2fbdd0716025e92bb3f6736abb36d 
    favor es6 destructuring

commit f8803d41d90ab4e3ca2085e3825eda11eaad04b9
    Remove extensions
```

PR #196, opened 2019-12-19
Pagination

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1154352301382233/f)
- EDS: Pagination for Data Grids

- [x] Navigation a11y
- [x] Adds "Nora Blue" per Tom's Figma spec
- [x] Incorporate Theinhardt per Tom's Figma spec

image


src/components/Grid/Pagination.js
```diff
@@ -0,0 +1,130 @@
...
+    })
+  }
+
+  const getPaginationNumbers = () => {
+    const pageNumbers = []
+    let paginationNumbers
+    if (pagingState.total !== null) {
+      for (
+        let i = 1;
+        i <= Math.ceil(pagingState.total / pagingState.per_page);
```
GreyFalcon: why is this logic needed when we have a variable `total_pages` ? 

Also, there's no check `per_page != null`.

src/components/Grid/Column.js
```diff
@@ -22,9 +22,12 @@ export const Column = React.memo((props) => {
    * tabIndex and point columnRef to itself so it gets the focus.
    */
   const interactiveRender = (props) => {
+    const klasses = props.header
```
lokeshy: I hope klasses is not typo . i like the variable name

src/components/Grid/Column.module.css
```diff
@@ -5,6 +5,12 @@
   box-sizing: border-box;
 }
 
+.columnheader {
+  text-transform: uppercase;
+  font-size: 11px;
```
lokeshy: Should we use em instead of px ? any thoughts how will px behave in responsive webpage .

src/components/Grid/Pagination.js
```diff
@@ -0,0 +1,130 @@
...
+    data,
+    total,
+    per_page,
+    total_pages,
+    current_page,
+    fetchPage,
+  })
+
+  const fetchPageDelegate = async (pageNumber) => {
+    const data = await fetchPage(pageNumber)
```
GreyFalcon: I might be late to the discussion,
but if the pagination experience lets the user choose the per page item count, that should also be part of the data sent to back-end, in addition to `pageNumber`.

Also, if there is a global sort (i.e., sort all the response data based on, say date) we need to pass in the `sortField` and `sortOrder` to the backend.
This is a longer discussion, but just putting my thoughts for the record.

src/components/Grid/Column.module.css
ankur-ethos: I am also curious to learn this. I actually like fixed font size, but we might want to consider the case when the text is too long.

src/components/Grid/Pagination.js
ankur-ethos: @GreyFalcon isn't `per_page` supposed to be the item count?

ankur-ethos: +1 On frontend providing the per page count.

ankur-ethos: @roblevintennis what is the plan for backend vs frontend sorting? Is it something that Tom needs to weigh in?

src/components/Grid/Column.js
stephaniel96: `klass` with a `k` is generally used instead of `class` to prevent reserved keywords from being used and to indicate that the variable is a class.  (https://stackoverflow.com/questions/4299289/what-is-the-difference-between-class-and-klass-in-ruby).

src/components/Grid/Pagination.js
stephaniel96: Just asked in our Slack channel about the two points above. I don't see a huge use case for building in the per page item count, but since I know that sorting was already something that Tom has in his mocks, `sortField` and `sortOrder` makes sense

src/components/Grid/Column.module.css
roblevintennis: Yes, this is a first iteration but we'll eventually need to add something like:

```css
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

src/components/Grid/Pagination.js
roblevintennis: Ha, yes, you're correct! I think I had it in before I elected to add the `total_pages` and also hadn't got buy in for that, but I'll switch to using `total_page` here 👍 

For the null check 1. I'll be switching to `total_pages` so it'll be a bit of a moot point, but, also, I think the server must always return `per_page` so the check for `pagingState.total` on line 41 makes it safe to assume an http response happened, and since the contract is to always supply that I don't know how rigorous we need to check for everything 🤔

src/components/Grid/Pagination.js
roblevintennis: > lets the user choose the per page item count,

Excellent feedback Ashwin!

I think for MVP we are not letting the user choose 10/20/50, but, the API should certainly allow for it I agree! It's hard to imagine we won't eventually offer that. Personally, I use that to enlarge the count per page whenever possible -- so I'm a fan!

Ultimately, I believe we should offer `page` and `per_page` so I agree.

From lack of push-back, I'm taking it that the format I've been asking for is fine with all. To be repetitive, it's something llike:

...json
page": 2,
    "per_page": 6,
    "total": 12,
    "total_pages": 2,
    "data": [...]
...
```

___

> what is the plan for backend vs frontend sorting

I think we can certainly be iterative, but, my first assumption would be:

- field sorting is passed in `field=direction` so `sortField` and `sortOrder` only need to be one query param: `?status=desc` 
- per Emma's input, `date=desc` will always be implied such that:
  1. default with no sort asked for will be sorted by date most recent at top aka
> Reverse chronological order would list them from latest to earliest.
  2. if we've asked for `?status=desc` that would be the primary sort, and if there are duplicates status, those will be secondarily ordered in reverse chronological ordering such that the most recent ones be listed before earlier ones

File: package.json [hunk 5]
```diff
@@ -31,6 +31,9 @@
   },
   "jest": {
     "verbose": true,
+    "setupFilesAfterEnv": [
+      "./setupTests.js"
```
roblevintennis: I needed to wire all this up to write my specs else I was getting regenerator-runtime/runtime related errors

src/components/Grid/Row.module.css
```diff
@@ -42,22 +42,14 @@
...
-  --default-row-hover-background: #deebff;
+  --default-row-hover-background: var(--BrandDuckegg);
 
   background-color: var(
     --ariagrid-row-hover-background,
     --default-row-hover-background
   );
 }
 
-.row:not(:first-child):hover > div:first-child {
```
roblevintennis: I was looking at Zendesk's table when I did this since Tom likes those. This CSS put a left border on the row when hovered, but was "too much like" their tables and so I removed it :)

PR #197, opened 2019-12-20
Adds ibm/plex

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1137984670709817/1146751046397448/f)
- EDS: Monospace Font

### DIFF
--- a/src/components/Type/Type.scss
+++ b/src/components/Type/Type.scss
@@ -2,6 +2,7 @@
    ========================================================================== */
 
 @import '../Media/Media.scss';
+@import '@ibm/plex/scss/ibm-plex.scss';
 
 // Typefaces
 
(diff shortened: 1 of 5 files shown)


roblevintennis:
I'm not sure why, but I wasn't able to actually get the font to render successfully, so I feel like this PR is still missing something but not sure what

zach-ethos:
Looking at the README, it doesn't look like importing the module magically works – I think it just installs assets, and you have to deeplink to them.

zach-ethos:
Oh wait, you're already doing that (with the .scss). Hm.

zach-ethos:
Did you try the .css? Not sure how to read this: https://github.com/ibm/plex#web-usage

roblevintennis:
The file is `@ibm/plex/scss/ibm-plex.scss` so I'm not sure what you mean by try the CSS. I see in the verbiage it says that, but the file referenced is clearly scss. In any event, I just don't see the font in the bottom of computed styles and looking at it against the correct one it definitely had different stems and serifs so I didn't get it to actually work.

PR #198, opened 2019-12-20
Adding a `<Portal />` Component

bwalker8038:
## Description
This PR adds the `<Portal />` component to the `EDS`. The `<Portal />` component.

The Portal component provides a way to create React portals dynamically. The component will create a new DOM node that attaches to the end of the document.body and render's the passed children within that DOM node.

This provides us a base for building Modals, Notifications, etc.

## Screenshots
Screen Shot 2019-12-20 at 2 30 57 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .vscode/
+.nova/
 
 node_modules
 yarn-error.log
--- a/src/components/Portal/Portal.js
+++ b/src/components/Portal/Portal.js
@@ -1,0 +1,68 @@
+import { useEffect, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+import PropTypes from 'prop-types'
+
+/**
+ *
+ * Automatically handles creating and tearing-down the root elements
+ * so there is no need to ensure the parent target already
+ * exists.
+ *
+ * @private
+ *
+ * @param {String} id - The id of the target container
+ *
+ * @return {HTMLElement} - The DOM node to use as the Portal target
+ */
+function usePortal(id) {
+  const [isAttached, setIsAttached] = useState(false)
+  const el = useRef(
+    document.getElementById(id) || document.createElement('div')
+  )
+
+  useEffect(() => {
+    setIsAttached(!!el.current.parentElement)
+
+    if (!isAttached) {
+      el.current.id = id
+      document.body.appendChild(el.current)
+      setIsAttached(true)
+    }
+
+    return () => {
+      if (isAttached) {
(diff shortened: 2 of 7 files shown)


src/components/Portal/Portal.js
```diff
@@ -0,0 +1,66 @@
...
+
+    if (!isAttached) {
+      el.current.id = id
+      document.body.appendChild(el.current)
+      setIsAttached(true)
+    }
+
+    return () => {
+      if (isAttached) {
+        el.current.parentElement.removeChild(el.current)
```
roblevintennis: Cleanup ftw! 🙌

src/components/Portal/Portal.js
```diff
@@ -0,0 +1,66 @@
...
+import { createPortal } from 'react-dom'
+import PropTypes from 'prop-types'
+
+/**
+ * Hooks handles the creation and tear down of the portal's root
+ * element.
+ *
+ * @private
+ *
+ * @param {String} id - The id of the target container
```
roblevintennis: Should maybe mention it's optional and that if not supplied a generic div will be created

src/components/Portal/Portal.js
```diff
@@ -0,0 +1,66 @@
...
+  const el = useRef(
+    document.getElementById(id) || document.createElement('div')
+  )
+
+  useEffect(() => {
+    setIsAttached(!!el.current.parentElement)
+
+    if (!isAttached) {
+      el.current.id = id
+      document.body.appendChild(el.current)
```
roblevintennis: Do we always want it attached to the `body` or should we accept an optional `container` element since you can ultimately do: `ReactDOM.createPortal(child, container)` and fallback default to `document.body`?

bwalker8038: So I think for dynamic portals we always want to attach to the `body`, which is the specific use case that this component solves. There might be a use case where we the `root` element is already rendered in the `DOM` (as static `html`), however I think in the majority of our cases, we want the `portal` to be created dynamically.

roblevintennis: Ok, well I've not used portals and so let's just go with what you've got here as you've been using this for real use cases and I trust you're more familiar with realistic usage. Thanks for clarifications Brad.

.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 .vscode/
+.nova/
```
zach-ethos: What's this?

bwalker8038: oh editor configs: https://panic.com/nova/

PR #200, opened 2019-12-24
Css modules with composes use scss extension

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1152334753103228/1154987346183273/f)

When I did a monorepo I bumped the version of EDS up to 1.1.41 and ran 
`yarn deploy:dryrun`

I got a few errors due to two things:

1. several `.module.css` files that confused monorepo because they're expecting to be used as CSS modules but monorepo's webpack configuration isn't as sophisticated as the EDS one (see this [Asana ticket](https://app.asana.com/0/1137984670709817/1154995542004083/f) for details on why and so the `composes:` clauses would error.
1. react-icons was missing in package.json. I'm pretty sure I must have inadvertantly removed that from package.json (yet it was in the yarn.lock). This adds it in.

- https://github.com/getethos/ethos/pull/1897

**Screenshots:**

image


src/components/Grid/useGridSorting.module.scss
```diff
@@ -1,6 +1,6 @@
 /**
  * Sort icons UX (from default, up, to down) seem to follow the same
- * conventions as Bootsprap: https://examples.bootstrap-table.com/
+ * conventions as Bootstrap: https:\/\/examples.bootstrap-table.com/
```
roblevintennis: This gets around a script we have that checks for improper use of `//` in `.module.scss` files

src/components/Grid/Grid.md
```diff
@@ -26,7 +26,7 @@ import { Grid } from './Grid.js'
 import { Row } from './Row.js'
 import { Column } from './Column.js'
 import { useGridSorting } from './useGridSorting.js'
-import styles from './grid-example.module.css'
+import styles from './grid-example.module.scss'
```
roblevintennis: Fwiw, I elected to just rename all `src/components/Grid/*` css files to .scss for consistency sake even though there were only a couple using `composes:` and confusing monorepo. My thought was it was better within this directory at least to be consistent

PR #201, opened 2019-12-26
[QA] Tooltip

tkarssli:
**Description:**
- [Asana Task]()
- Create a tooltip

- 

**Screenshots:**
image
image
image
image


package.json
```diff
@@ -40,10 +40,13 @@
     }
   },
   "dependencies": {
+    "@reach/dialog": "^0.7.0",
```
tkarssli: https://reacttraining.com/reach-ui/dialog/ 

Modals have a lot of accessibility features we need that I didn't think we needed to spend dev time creating so I found this lib which focuses on compatibility/accessibility and isn't bloated.

package.json
```diff
@@ -40,10 +40,13 @@
...
     "dayjs": "^1.8.16",
     "js-cookie": "^2.2.1",
     "lazysizes": "^5.1.2",
+    "react-popper": "^1.3.7",
```
tkarssli: Same lib we use on monorepo

package.json
```diff
@@ -40,10 +40,13 @@
...
     "react-select": "^3.0.8",
+    "react-spring": "^8.0.27",
```
tkarssli: https://www.react-spring.io/

Very well supported Lib that makes animations extremely easy, treeshakeable and relatively small. Thinking about future animations on EDS components that this could make a lot easier.

package.json
zach-ethos: cc @bwalker8038

package.json
zach-ethos: +1. I remember doing a deep dive on React animation a couple years ago and I recall this being one of the good ones. Curious what's planned for it. Did you look into other libs?

roblevintennis: Non-blocking

> 10.7KB (web), 9.7KB (react-native) or 6.9KB (/universal export). The size will ultimately depend on your build-chain and can decrease with tree-shaking. For instance, just importing useSpring without color-interpolations (using the /universal export) you will get 4.7KB.

I'd like to keep an eye out for how we use this and better understand it. Sounds like it's likely a good one, but let's try to sync up so I can stay in the loop.

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,199 @@
...
+  const [modalVisible, setModalVisibility] = useState(false)
+
+  const animationConfig = {
+    tension: 250,
+    clamp: true,
+  }
+
+  const transitions = useTransition(modalVisible, null, {
+    config: animationConfig,
+    from: { opacity: 0, transform: 'translate3d(0px, -20px, 0px)' },
```
roblevintennis: I wonder if the pixel values can be sync'd with our css `--space-N` ones? That would be ideal

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+      display: none;
+    }
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    margin: 12px;
+    text-align: center;
+    width: max-content;
+    max-width: 270px;
```
roblevintennis: This should probably be a bit more configurable

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+
+  .contentBox {
+    @include for-phone-only {
+      display: none;
+    }
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    margin: 12px;
```
roblevintennis: we need to use `--Space-N`

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+    text-align: center;
+    width: max-content;
+    max-width: 270px;
+    background-color: white;
+    font-size: 12px;
```
roblevintennis: There may be a default font def that we can use instead if you don't mind can you look for that and use it instead? Thanks sir!

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+    padding: 20px var(--Space-24) var(--Space-24) var(--Space-24);
+    box-shadow: 0px 0px 12px 5px #f1f1f1f1;
```
roblevintennis: can we use a standardized color and achieve similar affect?

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+    color: black;
```
roblevintennis: let's use the black we have defined

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+    transition: 0.1s ease-in-out;
```
roblevintennis: as we introduce animations, I think we should standardize on something like `--duration-short` or whatever e.g. 100 ms might be short 200ms might be default etc. etc.

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+
+    &.hidden {
+      opacity: 0;
+      visibility: hidden;
+    }
+  }
+
+  .arrow {
+    position: absolute;
+    width: 12px;
```
roblevintennis: spacing

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+        border-color: transparent white transparent transparent;
+      }
+    }
+    &[data-placement*='left'] {
+      right: 0;
+      margin-right: -12px;
+      height: var(--Space-24);
+      width: 12px;
+      &::before {
+        border-width: 12px 0 12px 12px;
```
roblevintennis: I realize it doesn't make sense to litter all places, but I think we need spacing for all of these. If we don't have something like 12px divisions we should think that through and either bump it to next value, or figure out a way to add a new --Spacing and/or speak with Design dept

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+      margin: auto;
+      display: block;
+      width: 0;
+      height: 0;
+      border-style: solid;
+    }
+  }
+}
+
+@keyframes fadein {
```
roblevintennis: Can we centralize all keyframes for reuse?

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,167 @@
...
+}
+
+.mobileModal {
+  padding: var(--Space-64) var(--Space-24) var(--Space-80) var(--Space-24);
+
+  .closeButton {
+    background: transparent;
+    border: unset;
+    position: absolute;
+    top: 29px;
```
roblevintennis: 29 seems a bit odd here. Is this making something fit flush against another box or something?

tkarssli: I had pulled it out of the figma but it might be a mistake? 
image

src/components/Tooltip/Tooltip.module.scss
tkarssli: Removed!

package.json
bwalker8038: @tkarssli – I'm currently working on a PR to add modals into the EDS that will be up this week.

src/components/Tooltip/Tooltip.module.scss
roblevintennis: I'd push back and/or verify with design :) Ideally we'd be using `32px` or some multiple consistent with our Spacing api

tkarssli:
fix z-indexing, outlines on xbutton

roblevintennis:
@bwalker8038 I spoke with @tkarssli today and we're wondering if it makes sense for him to pull this through tomorrow as a first iteration, and then later, when you've got the modal stuff done, replace that stuff with yours in a second iteration. Any thoughts on the feasibility of doing it that way? Also, is end of week for the modal realistic?

src/components/Tooltip/Tooltip.js
tkarssli: Synced and created  `--UncommonSpace-N` vars for situations like this

src/components/Tooltip/Tooltip.module.scss
tkarssli: Fixed

src/components/Tooltip/Tooltip.module.scss
tkarssli: Used `--GrayLightHover--opaque`

src/components/Tooltip/Tooltip.module.scss
tkarssli: Changed to `var(--GrayDarkHover--translucent)`

src/components/Tooltip/Tooltip.module.scss
tkarssli: changed to use ` --transition-default`

src/components/Tooltip/Tooltip.module.scss
tkarssli: Change to use a new uncommone spacer

src/components/Tooltip/Tooltip.module.scss
tkarssli: I changed this to use an uncommon spacing Variable since this is a unique case where we need 12px;

src/components/Tooltip/Tooltip.module.scss
tkarssli: Remove keyframes

src/components/Tooltip/Tooltip.module.scss
tkarssli: Fixed this

src/components/Tooltip/Tooltip.js
tkarssli: Actually wait this is in in the `.js` I am not sure how to do this

src/components/Spacer/Spacer.scss
```diff
@@ -25,4 +25,21 @@
   --NegativeSpace-16: -16px;
   --NegativeSpace-8: -8px;
   --NegativeSpace-4: -4px;
+
+  --UncommonSpace-2: 2px;
```
roblevintennis: Hrm, I'm not sure I'm a fan of this many of these and/or if it'd just be better to leave it hard coded with the `12px` places cc @zach-ethos any thoughts?

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,285 @@
+import React, { useState, useEffect, useRef, Fragment } from 'react'
+import PropTypes from 'prop-types'
+import { DialogOverlay, DialogContent } from '@reach/dialog'
```
roblevintennis: @bwalker8038 is your modal done? I think this component should probably just use your modal if so

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,285 @@
...
+            )}
+          </Popper>
+        </div>
+      </Manager>
+    </div>
+  )
+
+  return (
+    <>
+      {isMobile() && renderModal}
```
roblevintennis: It just occurred to me that we don’t have any test coverage for the tooltip 🤔

package.json
tkarssli: Removed these libs for now

package.json
tkarssli: Removed this lib for now

src/components/Tooltip/Tooltip.js
tkarssli: Removed this transition

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,274 @@
...
+        data-placement={placement}
+      />
+    </div>
+  )
+}
+
+Tooltip.SVGS = {
+  closeButton: (
+    <svg
+      width="14"
```
roblevintennis: Nit: I usually prefer to set this via CSS ... but we don't have a proper SVG pipeline 🤦‍♂ so not your fault/issue here

PR #202, opened 2019-12-27
Typescript typings for grid stuff. Also, I was missing the Pagination…

roblevintennis:
No Asana...this is counterpart to https://github.com/getethos/ethos/pull/1903 which requires the Data Grid and Pagination stuff is exported and also Typescript typed.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.42",
+  "version": "1.1.45",
```
roblevintennis: The reason for bump from 42 to 45 is I've found yarn link and local development flaky, so I just prefer to bump the EDS patch versions until it's working properly ;)

src/components/index.js
```diff
@@ -10,7 +10,7 @@ export { COLORS } from './Colors'
 export { EmailInput } from './EmailInput'
 export { Faq } from './Faq'
 export { Form } from './Form'
-export { Grid, Row, Column, useGridSorting } from './Grid'
+export { Grid, Pagination, Row, Column, useGridSorting } from './Grid'
```
ankur-ethos: is this an unused import?

src/components/index.d.ts
```diff
@@ -286,6 +286,50 @@ export declare namespace RadioButtonGroup {
...
+export declare const useGridSorting: (
+  rows: any,
+  columns: any
+) => {
+  rowsRefs: any
+  columnRefs: any[]
+  sortedRows: any
+  compareBy: (
+    key: any,
+    sortMethod?: (a: any, b: any) => 1 | 0 | -1
```
ankur-ethos: what's the default implementation of `sortMethod`?

src/components/index.d.ts
```diff
@@ -286,6 +286,50 @@ export declare namespace RadioButtonGroup {
...
+    renderCallback: any
+  }): JSX.Element
+  propTypes: {
+    fetchPageCallback: any
+    renderCallback: any
+  }
+  displayName: string
+}
+export declare const useGridSorting: (
+  rows: any,
```
ankur-ethos: just for my understanding, do these really need to be `any`?

roblevintennis: Probably not but I plan to be iterative and refine these types further

src/components/index.js
roblevintennis: It’s an export that is now needed now that I wanted to consume it from outside of EDS.

roblevintennis:
Thanks guys. I agree we can do better in terms of typing and I definitely plan to continue improving in the area of EDS typings. I’m gonna merge this in spirit of being iterative and continue to refine over next couple/few weeks.

PR #204, opened 2019-12-31
Stronger types for CheckboxInput

roblevintennis:
**Description:**
- CheckboxInput typings

Verification

**Screenshots:**
image

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -30,30 +30,31 @@ export declare const CheckboxInput: {
     ...rest
   }: {
     [x: string]: any
-    formChangeHandler: any
-    validator: any
+    formChangeHandler?: any
+    validator?: any
     children: any
-    disabled: any
-    name: any
-    initialValue: any
-    currentValue: any
-    currentError: any
-    setFieldTouched: any
-    formTouched: any
+    disabled?: boolean
+    name: string
+    'data-tid': string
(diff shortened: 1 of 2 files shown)


src/components/index.tests.tsx
```diff
@@ -13,8 +16,26 @@ const demoQuestions = [
...
     return <Faq questions={demoQuestions} />
   }
 }
+
+class CheckboxInputTest extends React.Component<any, any> {
+  render() {
+    return (
+      <>
+        <CheckboxInput
+          name="name: string -- is required"
```
roblevintennis: My main goals on these iterations are:

1. stronger typing
2. allow for optional props e.g. `foo?: TYPE`

I noticed with the Nora prototyping stuff that some of the EDS components in that env were requiring all props to be supplied even if with `prop={undefined}` which  is silly.

So, my thought is testing with only required props helps to validate that here

src/components/index.d.ts
```diff
@@ -30,30 +30,31 @@ export declare const CheckboxInput: {
     ...rest
   }: {
     [x: string]: any
-    formChangeHandler: any
-    validator: any
+    formChangeHandler?: any
+    validator?: any
```
roblevintennis: On later iterations I'll deal with typing functions as I'll probably want to create interfaces and what not.

PR #209, opened 2020-01-01
Typing improvements email input

roblevintennis:
**Description:**
- EmailInput typings

**Verification**


src/components/EmailInput/EmailInput.js
```diff
@@ -43,7 +43,7 @@ EmailInput.propTypes = {
   disabled: PropTypes.bool,
   allCaps: PropTypes.bool,
   name: PropTypes.string.isRequired,
-  labelCopy: PropTypes.string.isRequired,
```
roblevintennis: We have the following default so `isRequired` really doesn't make sense here

```js
EmailInput.defaultProps = {
  labelCopy: 'Email',
...
```

PR #211, opened 2020-01-02
TextMaskedInput typing improvements

roblevintennis:
**Description:**
- TextMaskedInput typings

**Verification**

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -373,40 +373,40 @@ export declare const EmailInput: {
 export declare const TextMaskedInput: {
   (props: any): JSX.Element
   PUBLIC_PROPS: {
-    doValidation: any
-    placeholder: any
     mask: any
-    guide: any
-    initialValue: any
-    keepCharPositions: any
-    pipe: any
-    type: any
-    'data-tid': any
-    disabled: any
-    allCaps: any
-    name: any
-    labelCopy: any
-    validator: any
-    setTouched: any
-    getTouched: any
+    type: string
+    'data-tid': string
+    name: string
+    labelCopy: string
+    doValidation?: any
+    placeholder?: any
+    guide?: boolean
+    initialValue?: string
+    keepCharPositions?: boolean
+    pipe?: any
+    disabled?: boolean
+    allCaps?: boolean
+    validator?: any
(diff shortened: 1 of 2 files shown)


src/components/index.d.ts
```diff
@@ -373,40 +373,40 @@ export declare const EmailInput: {
...
+    type: string
+    'data-tid': string
+    name: string
+    labelCopy: string
+    doValidation?: any
+    placeholder?: any
+    guide?: boolean
+    initialValue?: string
+    keepCharPositions?: boolean
+    pipe?: any
```
bwalker8038: If this is a function, the type signature should be something like this:

```ts
  pipe?: () => void // or any other return type: any, T (Generic), boolean, etc)
```

src/components/index.d.ts
```diff
@@ -373,40 +373,40 @@ export declare const EmailInput: {
...
+    disabled?: boolean
+    allCaps?: boolean
+    validator?: any
```
bwalker8038: see above

src/components/index.d.ts
```diff
@@ -373,40 +373,40 @@ export declare const EmailInput: {
...
-    labelCopy: any
-    validator: any
-    setTouched: any
-    getTouched: any
+    type: string
+    'data-tid': string
+    name: string
+    labelCopy: string
+    doValidation?: any
+    placeholder?: any
```
bwalker8038: Nit: this should be a `string`

src/components/index.d.ts
```diff
@@ -373,40 +373,40 @@ export declare const EmailInput: {
...
-    name: any
-    labelCopy: any
-    validator: any
-    setTouched: any
-    getTouched: any
+    type: string
+    'data-tid': string
+    name: string
+    labelCopy: string
+    doValidation?: any
```
bwalker8038: If this is a function, the type signature should be something like this:

```ts
  pipe?: () => void // or any other return type: any, T (Generic), boolean, etc)
```

roblevintennis: Thanks @bwalker8038 -- I was going to save the functions for another pass, but I think I'll just go ahead and start on a follow up now!

PR #212, opened 2020-01-02
ZipInput typing improvements

roblevintennis:
**Description:**
- ZipInput typings

**Verification**

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -475,18 +475,18 @@ export declare const InfoMessage: {
 export declare const ZipInput: {
   (props: any): JSX.Element
   PUBLIC_PROPS: {
-    'data-tid': any
-    labelCopy: any
-    name: any
+    'data-tid': string
+    labelCopy: string
+    name: string
     disabled?: boolean
     allCaps?: boolean
     validator?: any
     initialValue?: string
   }
   propTypes: {
-    'data-tid': any
-    labelCopy: any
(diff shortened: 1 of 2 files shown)


src/components/index.d.ts
```diff
@@ -476,21 +476,21 @@ export declare const ZipInput: {
   (props: any): JSX.Element
   PUBLIC_PROPS: {
     'data-tid': any
-    disabled: any
-    allCaps: any
-    name: any
     labelCopy: any
-    validator: any
-    initialValue: any
+    name: any
```
jameshawkinsjr: It looks like the propTypes for `data-tid`, `labelCopy`, and `name` are `PropTypes.string.isRequired` - is there a reason why we'd use `any` here?

src/components/index.d.ts
```diff
@@ -476,21 +476,21 @@ export declare const ZipInput: {
...
   }
   propTypes: {
     'data-tid': any
-    disabled: any
-    allCaps: any
-    name: any
     labelCopy: any
-    validator: any
-    initialValue: any
+    name: any
```
jameshawkinsjr: Same Q as above

src/components/index.d.ts
roblevintennis: silly mistake -- fixed in 9d92c639b2d0d495b587646007e327ebcbb78876 thanks for catching

PR #214, opened 2020-01-02
Typing for NumberInput. But, also went ahead and did a bunch of dupli…

roblevintennis:
**Description:**
- NumberInput typings
- While I was in here, I elected to also update the signatures for `formChangeHandler` and `validator` methods across the full index.d.ts

**Verification**

### DIFF
--- a/src/components/index.tests.tsx
+++ b/src/components/index.tests.tsx
@@ -3,6 +3,7 @@ import { ButtonSelectGroup } from './index'
 import { CheckboxInput } from './index'
 import { EmailInput } from './index'
 import { Faq } from './index'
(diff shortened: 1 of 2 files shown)


src/components/index.tests.tsx
```diff
@@ -13,6 +14,35 @@ import { ZipInput } from './index'
 
 // -------------- Localized tests ----------------//
 
+class ButtonSelectGroupTest extends React.Component<any, any> {
```
roblevintennis: I just moved ButtonSelectGroupTest and CheckboxInputTest up to be in alphabetical order

PR #216, opened 2020-01-03
Button and children: React.ReactNode

roblevintennis:
**Description:**
- Button
- `children: React.ReactNode` (from `:any`)

- 

**Screenshots:**

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -74,14 +74,29 @@ export declare const Logo: {
   Inline: typeof LogoInline
 }
 
+interface downstreamProperties {
+  backArrowIcon?: boolean,
+  arrowIcon?: boolean,
+  type: string,
+  isSelected: boolean,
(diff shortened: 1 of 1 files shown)


src/components/index.d.ts
```diff
@@ -74,14 +74,29 @@ export declare const Logo: {
...
+interface downstreamProperties {
+  backArrowIcon?: boolean,
+  arrowIcon?: boolean,
+  type: string,
+  isSelected: boolean,
+  fullWidth: boolean,
+  disabled: boolean,
+  name: string,
+  onClick: any,
+  'data-tid': string,
```
GreyFalcon: nit: any reason to make this as string?
(Not blocking, just a curiosity question)

roblevintennis: Ethos codes has always had these data tids as data attributes with string values used for making tests against so it's of type `:string`

PR #217, opened 2020-01-03
A bit of TLC to the Grid typings.

roblevintennis:
**Description:**
- Grid et al Typing

**Verification**

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -315,36 +315,50 @@ export declare namespace RadioButtonGroup {
     required?: boolean
   }
 }
-
 export declare const Grid: {
   (props: any): JSX.Element
   propTypes: {
     children: React.ReactNode
-    className?: any
+    className?: string
     rowRefs: any
     columnRefs: any
   }
   defaultProps: {
-    className: any
+    className: string
   }
   displayName: string
 }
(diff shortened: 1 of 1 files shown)


src/components/index.d.ts
```diff
@@ -315,36 +315,50 @@ export declare namespace RadioButtonGroup {
...
   ({
     fetchPageCallback,
     renderCallback,
   }: {
-    fetchPageCallback: any
-    renderCallback: any
+    /**
+     * Ultimately, this returns the JSON.parse'd data
+     */
+    fetchPageCallback: (pageNumber: number | string) => any
```
gbockus: should any -> object?

roblevintennis: Hrm, yeah, possibly huh. Did some googling and this was interesting: https://stackoverflow.com/a/58143592

I'll try adding this into https://github.com/getethos/ethos-design-system/pull/219 which is being consumed by Nora in monorepo and still a live PR :)

roblevintennis: Yah, it seemed to work fine! I added it into [this commit](https://github.com/getethos/ethos-design-system/pull/219/commits/a6a174529d8054a32a40d319804763c882963112) and here's the Nora example:

image

PR #218, opened 2020-01-03
Adds validation, errors, and field label to Select

wafflehorse:
**Description:**
- Supports this [Add DL edit functionality to user profiles in admin](https://app.asana.com/0/1149628124018183/1153703353529217)
- Validation for Forms was not hooked into our Select component. Also, labelCopy was not displayed like other form fields. This PR adds this stuff to Select. 
- For the future, it feels like we could use some sort of field wrapper to hold form-level logic as our components need a decent amount of form logic baked into them now. As you can see here, our simple Select component isn't so simple anymore. 

**Screenshots:**
Screen Shot 2020-01-03 at 3 25 27 PM

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 // See React-Select -- https://github.com/JedWatson/react-select for documentation
 // on usage, Async configuration, etc.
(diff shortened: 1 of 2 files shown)


src/components/Select/Select.js
```diff
@@ -1,22 +1,70 @@
...
+    if (isSelectedValue) {
+      validationSelect()
+    }
+  }, [selectedValue])
+
+  const onBlur = () => {
+    validationSelect()
+  }
+
+  const wrapperClass = title ? styles.wrapper : ''
```
roblevintennis: How is title tied to the wrapper styles

src/components/Select/Select.js
```diff
@@ -49,6 +99,12 @@ Select.propTypes = {
   title: PropTypes.string,
   className: PropTypes.string,
   isCreatable: PropTypes.bool,
+  formChangeHandler: PropTypes.func,
+  currentError: PropTypes.string,
+  formTouched: PropTypes.bool,
+  labelCopy: PropTypes.string,
```
roblevintennis: Should we make `labelCopy` required like many of the other components do? I think this could also make the related conditional rendering simplified as well in that you wouldn’t have to check it is defined

src/components/Select/Select.js
wafflehorse: This line of code wasn't something I added. I simply moved it. So I'm not sure...

src/components/Select/Select.js
wafflehorse: The reason why I didn't make it required is because the Select component has the title prop which is essentially a label that lives within the border of the select box. So you wouldn't want both the title and labelCopy, but I thought if you weren't using the select box in a form, you may want the title prop instead of labelCopy. I wasn't planning on adding labelCopy until I realized that it looks inconsistent with the other fields. But, it does make the component more confusing to use. What do you think?

bwalker8038: @roblevintennis I would agree with this. I think in the majority of form elements we have implemented in the `EDS`, the label is required and set in the component.

wafflehorse: Sounds good! I'm gonna go ahead and make labelCopy and name required.

wafflehorse: Scratch that ^ Leaving it flexible for now after discussion.

PR #219, opened 2020-01-03
EDS / Nora Issues

roblevintennis:
This PR fixes issues testing pulling in latest v1.1.46 EDS into Nora...

**Screenshots:**

Fixed in 6de837a9cc93af0fbce6cde68e451752805c59d1

image

Fixed in 70fb439d98f24b2a4a46af4a3ddfb60c2053e6f6 ... we needed to add `void` to the union type for `renderCallback`. Use case is we want to just set state with updated rows and have that trigger a re-render that picks up rendering the grid from another method:

```jsx
  const renderCases = (rows: any) => {
    if (!rows || !rows.length) {
      return
    }
    /*** Here we are returning void but also updating state which will ultimately ***/
    /*** kick in renderGrid method which renders the `JSX.Element | null` ***/
    setRows(rows) 
  }
  const renderGrid = () => {
    if (rows.length) {
      return <LeGrid rows={rows} columns={columns} /> 
    }
    return null
  }

  return (
    <>
      {renderGrid()}
      <Pagination fetchPageCallback={fetchPage} renderCallback={renderCases} />
    </>
  )
```

image

ac48fb9dd3786277d40367583b762079825d3c85 fixed a "not assignable to type" for `validateExists` ...I forgot a screenshot

a6a174529d8054a32a40d319804763c882963112 is an improvement per Gordon's recommendation on another PR to switch return type of `any` to `object`

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -292,7 +292,7 @@ export declare namespace RadioButtonGroup {
     currentValue?: string
     currentError?: string
     formChangeHandler?: (value: string, errorValue: string) => void
-    onChange?: any
+    onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void
     'data-tid'?: string
     validator?: (value: string) => string
     disabled?: boolean
@@ -308,7 +308,7 @@ export declare namespace RadioButtonGroup {
     currentValue?: string
     currentError?: string
     formChangeHandler?: (value: string, errorValue: string) => void
-    onChange?: any
+    onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void
     'data-tid'?: string
     validator?: (value: string) => string
     disabled?: boolean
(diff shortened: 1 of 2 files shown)


src/components/index.d.ts
```diff
@@ -74,36 +74,36 @@ export declare const Logo: {
...
   backArrowIcon?: boolean,
   arrowIcon?: boolean,
-  type: string,
-  isSelected: boolean,
-  fullWidth: boolean,
-  disabled: boolean,
-  name: string,
-  onClick: any,
-  'data-tid': string,
+  type?: string,
```
roblevintennis: Main fix was that I needed to make these optional

src/components/index.d.ts
```diff
@@ -336,21 +336,21 @@ export declare const Pagination: {
     /**
      * Ultimately, this returns the JSON.parse'd data
      */
-    fetchPageCallback: (pageNumber: number | string) => any
+    fetchPageCallback: (pageNumber: number | string) => object
```
roblevintennis: Relates to @gbockus feedback at https://github.com/getethos/ethos-design-system/pull/217#discussion_r362981617

roblevintennis:
Not sure I understand the question re safety measures for accidental pulls. Can you explain?

PR #220, opened 2020-01-06
Fixes initial values and validation for Select

wafflehorse:
**Asana** [https://app.asana.com/0/1136962714401929/1155914817079659/f] -- EDS Select Initial Values and Validation

**Description:**
- If you specify an initial value on the select box, then onblur without changing the value, validateExists, for example, would fail since the selectedValue is initially set to undefined instead of the passed in value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.48",
+  "version": "1.1.49",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -30,7 +30,7 @@ export const Select = ({
 }) => {
   const resolvedValidator = validator ? validator : () => ''
   const [selectedOption, updateSelectedOption] = useState(initialValue)
-  const [touched, setTouched] = useState(false)
+  const [touched, setTouched] = useState(initialValue ? true : false)
   const [getError, setError, , validate] = useErrorMessage(resolvedValidator)
   const [doValidation] = useInputValidation({
     validate,
@@ -42,7 +42,7 @@ export const Select = ({
     selectedOption ? selectedOption.value : undefined
 
   useEffect(() => {
-    if (!!formChangeHandler && touched) {
+    if (touched) {
       doValidation(getSelectedValue(), touched)
     }
   }, [selectedOption])
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -28,8 +28,10 @@ export const Select = ({
   options,
   ...rest
 }) => {
+  const defaultValue = options.find((option) => option.value === initialValue)
   const resolvedValidator = validator ? validator : () => ''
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.46",
+  "version": "1.1.47",
```
roblevintennis: I think this will soon have to be bumped up to .49 or something due to competing branches. Should be as easy as merging in lastest and running `yarn version` and then pushing up and also `git push --tags`

src/components/Select/Select.js
```diff
@@ -23,37 +24,43 @@ export const Select = ({
...
-    }
-  }
+  const getSelectedValue = () =>
+    selectedOption ? selectedOption.value : undefined
 
   useEffect(() => {
-    const isSelectedValue = typeof selectedValue !== 'undefined'
-    if (isSelectedValue) {
-      validationSelect()
+    if (!!formChangeHandler && touched) {
```
wafflehorse: @roblevintennis I cleaned this up a bit

package.json
wafflehorse: I'll do that now

src/components/Select/Select.js
```diff
@@ -23,37 +24,41 @@ export const Select = ({
...
+    if (onChange) {
+      onChange(event)
     }
-  }, [selectedValue])
+  }
 
   const onBlur = () => {
-    validationSelect()
+    setTouched(true)
+    doValidation(getSelectedValue(), true)
```
wafflehorse: @roblevintennis I noticed that in CheckboxInput we don't check if formChangeHandler exists before validating in onBlur like we do in useEffect. I copied that pattern here, but I don't know why we wouldn't need that here as well.

src/components/Select/Select.js
```diff
@@ -23,37 +24,41 @@ export const Select = ({
...
-    }
-  }
+  const getSelectedValue = () =>
+    selectedOption ? selectedOption.value : undefined
 
   useEffect(() => {
-    const isSelectedValue = typeof selectedValue !== 'undefined'
-    if (isSelectedValue) {
-      validationSelect()
+    if (!!formChangeHandler && touched) {
```
roblevintennis: not sure if the `useInputValidation` is doing that check for us already in which case this'd be duplication

src/components/Select/Select.js
roblevintennis: Yeah, I think the `doValidation` is coming from the `useInputValidation` hook which is doing it for us perhaps

src/components/Select/Select.js
wafflehorse: yeah, you're right! I removed the formChangeHandler, but I needed to keep the touched check because otherwise the form would be 'interacted with' immediately.

src/components/Select/Select.js
```diff
@@ -23,37 +24,41 @@ export const Select = ({
   formTouched,
   labelCopy,
   name,
+  initialValue,
```
roblevintennis: I'm not seeing `initialValue` take affect:

image

PR #222, opened 2020-01-09
[Nora] Tag Component

stephaniel96:
**Description:**
- [Asana Ticket](https://app.asana.com/0/1150068311310825/1155921313537879)
- Basic `Tag` component for Nora. Below are some examples from the mocks. Details, colors, and icons are still TBD. 
- I'm aware that the text is not vertically centered. This is due to the font, and will be fixed at a later date (I am assuming Tag colors and icons will change at least once before the final Nora product).

| Designs | EDS |
| --- | --- |
| image | image |

- 

**Screenshots:**


src/nora/components/Tag/Tag.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '../Attributes.css';
```
stephaniel96: @roblevintennis is this import necessary? I couldn't get `--Radius-3` to work without this explicit import, but didn't have to do this with the colors

src/nora/components/Colors.css
```diff
@@ -1,3 +1,12 @@
 :root {
   --NoraBlue: #1a7ff6;
+
+  --GreenPrimary--opaque: #2bd04b;
```
roblevintennis: INFO: I wonder if we should make these consistent with `NoraBlue` in terms of a naming convention? I realize the "Nora" part is a bit clumsy, but my thought was it would prevent folks from inadvertently using these outside of Nora.

src/nora/components/Tag/Tag.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from './Tag.module.scss'
+
+const Tag = ({ children, type }) => {
+  let colorClass
+  console.log(type)
```
roblevintennis: Prune the console.log :)

src/nora/components/Tag/Tag.js
```diff
@@ -0,0 +1,31 @@
...
+    case 'approved':
+      colorClass = styles.Approved
+      break
+    case 'rejected':
+      colorClass = styles.Rejected
+      break
+    default:
+      colorClass = styles.Default
+  }
+  console.log(colorClass)
```
roblevintennis: here too

src/nora/components/Tag/Tag.module.scss
roblevintennis: Yeah, I actually think it IS required so I think you're good the way you have it. Only way it wouldn't be is if we unconditionally were loading those definitions which we are not. I prefer module imports where possible anyway, so I like the pattern you're using here 👍

src/nora/components/Tag/Tag.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.Tag {
+  border-radius: var(--Radius-3);
+  display: inline;
+  align-items: center;
+  padding: 5px;
```
roblevintennis: It would be preferable to use `var(Space-<N>)` here I think we have 4px and 8px for that. I'm not sure if that's doable. If not, I'd 1) double check with Tom if it has to be this way (if this is just to match the design) 2) I suppose we could revisit this and just go with the hard coded 5px.

src/nora/components/Tag/Tag.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '../Attributes.css';
+
+.Tag {
+  border-radius: var(--Radius-3);
+  display: inline;
```
roblevintennis: Can this be `inline-block instead?

src/nora/components/Tag/Tag.module.scss
stephaniel96: 4 looks a little tight but I think it's okay for now! I'll stick to using the defaults and when we go through a fine-tune everything we can decide when to make exceptions.

src/nora/components/Tag/Tag.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from './Tag.module.scss'
+
+const Tag = ({ children, type }) => {
+  let colorClass
+  switch (type) {
+    case 'approved':
```
roblevintennis: Pretty sure we'll need some more states, but this is good and we'll be iterative! 🙌

PR #225, opened 2020-01-17
Conditionally paging buttons and add labelCopy prop to grid

roblevintennis:
**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1155921313537887/f)
- Pagination should hide itself if items is <= number of items per page
- Data Grid column headers need to support `labelCopy` property

- https://github.com/getethos/ethos/pull/2012

**Screenshots:**

# Hiding Pagination

Context: We noticed on the Nora side that if you only have a single page of results, the pagination arrow buttons still showed up which is a poor UX...

Consuming from Nora example with 6 items (1 page). Thing to note is the pagination is now completely hidden since there's only 1 page.

image

Nominal case (there are pages to paginate) still works:

image

# labelCopy

Context on the issue: 
image

So we were using the `col.name` for both key and label which is of course silly 😋 

___

Turned out nothing needed to be added in terms of incoming component props. We just need to use dataset and utilize the key and label copy properly. By design, the data grid inverts control to the consumer in how they want to build up the cell to allow for more flexbility, and then ultimately just renders that `child` or `children` if you will.

The work done was to update the examples and encourage use of `labelCopy` with an optional approach of using the `name` as a fallback if you'd like.

Here's the EDS example:

image

And here's the Nora EDS Fetch Applications example (after consuming these EDS updates):

image

### DIFF
--- a/src/components/Grid/Grid.md
+++ b/src/components/Grid/Grid.md
@@ -1,4 +1,5 @@
 ### Accessible React Grid
+
 This `Grid` follows the [wai-aria-practices-1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#grid) specifications for data grids. It's influenced by [juanca's](https://github.com/juanca) original [react-aria-components](https://github.com/juanca/react-aria-components) data-grid component in the way the keyboard navigation and roles are assigned, but otherwise is a complete rewrite with the following added features:
 
 - The class-based lifecycle components were all converted to use functional hook-based components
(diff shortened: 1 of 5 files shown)


src/components/Grid/Pagination.js
```diff
@@ -14,32 +14,42 @@ export const Pagination = memo(({ fetchPageCallback, renderCallback }) => {
     fetchPage: fetchPageCallback,
   })
 
+  const conditionallyRenderPagingButtons = () => {
```
zach-ethos: `conditionallyDoThing` sounds so much better than `maybeDoThing` which Eric gave me a hard time about once

src/components/Grid/Grid.md
```diff
@@ -211,7 +218,7 @@ function GridExample() {
   )
 }
 
-<GridExample />
+;<GridExample />
```
gbockus: why do we do this?

src/components/Grid/Pagination.js
gbockus: `couldDoThingDontWorryAboutIt`

src/components/Grid/Grid.md
roblevintennis: It's an artifact of react styleguidist as it sometimes gets confused if there's previous stuff before like imports and what not. I think it's actually a dependency of react-styleguidist maybe `acorn-jsx` 🤔

PR #227, opened 2020-01-18
Nora Header component

roblevintennis:
…d total wip

**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1157190385219581/f)
- Nora Header component

**Screenshots:**

After refactoring to take left children and right children props instead of the less flexible Title and Profile. Also, it adds `<nav>` which feels more semantic after doing a bit of research:

image

JSDocs:

image

### DIFF
--- a/src/nora/components/Colors.css
+++ b/src/nora/components/Colors.css
@@ -10,4 +10,5 @@
   --NoraGray--opaque: #4f5972;
   --NoraGray--translucent: #eaebee;
   --NoraGrayHeader--opaque: #7c8699;
+  --NoraGrayHeaderBar--opaque: #dfe1e6;
 }
--- a/src/nora/components/Colors.css
+++ b/src/nora/components/Colors.css
@@ -9,4 +9,5 @@
 
   --NoraGray--opaque: #4f5972;
   --NoraGray--translucent: #eaebee;
+  --NoraGrayHeader--opaque: #7c8699;
 }
--- a/src/nora/components/Example/Example.js
+++ b/src/nora/components/Example/Example.js
@@ -1,5 +1,5 @@
-import React from 'react'
 import PropTypes from 'prop-types'
+import React from 'react'
 import styles from './Example.module.scss'
 
 const INFO_MESSAGE_TYPES = {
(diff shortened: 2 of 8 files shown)


src/nora/components/Colors.css
```diff
@@ -9,4 +9,6 @@
 
   --NoraGray--opaque: #4f5972;
   --NoraGray--translucent: #eaebee;
+  --NoraGrayHeader--opaque: #7c8699;
```
roblevintennis: I've sent an email regarding "Nora Gray's", so I think this will need further iterating. But it's fine for now I think.

src/nora/components/Example/Example.js
```diff
@@ -1,5 +1,5 @@
-import React from 'react'
 import PropTypes from 'prop-types'
+import React from 'react'
```
roblevintennis: My setup is reordering these

src/nora/components/Header/Header.js
```diff
@@ -0,0 +1,67 @@
...
+ * @private
+ *
+ * @param {JSX.Element} props.children - left children
+ */
+const Right = ({ children }) => {
+  return <div className={styles.Right}>{children}</div>
+}
+
+const Navbar = ({ children }) => {
+  return <nav className={styles.Navbar}>{children}</nav>
```
roblevintennis: After doing a bit of research, doing `<header><nav>...</nav></header>` feels most semantic to me.

src/nora/components/Example/Example.js
stephaniel96: Interesting. We should see if we should codify this in a linting rule. Otherwise, down the road the git blame and PR diffs might get confusing. 

I’m in favor of grouping external and internal imports, but don’t have a particular opinion of ordering within that

roblevintennis: Good point Steph. So, I figured it out. It's `source.organizeImports` 

```json
"editor.codeActionsOnSave": {
  "source.organizeImports": true,
  "source.fixAll": true
}
```

I think I'll just remove it and place this back. I feel like it might be more idiomatic to have React at the top...I'll push up another commit once I do.

stephaniel96: Awesome thanks for digging into it Rob!

PR #228, opened 2020-01-19
Docker

roblevintennis:
No Asana yet. This is an experiment with Dockerizing EDS


package.json
```diff
@@ -35,6 +35,7 @@
       "./setupCoreJsPollyfills.js"
     ],
     "testPathIgnorePatterns": [
+      "<rootDir>/src/components/index.test.tsx",
```
roblevintennis: This is the test file for typescript bindings and Jest will choke on it if ran. I had it as `index.tests.tsx`, but then found when I tried to run styleguidist through docker, it chokes due to no proper loader. Not sure why it doesn't happen locally, but it made sense to rename it so styleguidist would automatically `ignore` as `.test` files are ignored by default.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Docker cheatsheet
```
roblevintennis: I usually wouldn't comment so pedantically, but as a docker noob, I feel this will help me for a while 😛

src/components/tsconfig.json
```diff
@@ -10,5 +10,5 @@
     "noEmit": true,
     "esModuleInterop": true
   },
-  "files": ["index.d.ts", "index.tests.tsx"]
+  "files": ["index.d.ts", "index.test.tsx"]
```
roblevintennis: React Styleguidist will ignore .test but not .tests:

https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#ignore

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Install node modules
+COPY ./package.json ./
+RUN yarn install
+
+# The following assumes you've already pulled in the fonts to your local
+# EDS checkout. If not, download `fonts.zip` from:
+# [Google Drive](https://drive.google.com/drive/u/0/folders/1hvAAEUWEsz2Hq-Wmj09xOvCYeixkJ4_c)
+# and have unzip'd it into EDS's `src/fonts` directory
+COPY ./ ./
```
zach-ethos: 🤔

roblevintennis: this copies over the entire design dir local --> docker container. I think for EDS this is cleaner then curating over specific thing which would add to the complexity and hurt readability. Do you disagree?

alexthecarp:
I'm ok with copying everything actually. It's simple to reason about. If
the concern is size, we can do multi stage build

-- 
Alex Carp

roblevintennis:
I'm honestly wondering what I should do with this. I don't want to succumb to ["speculative generality"](https://blog.codinghorror.com/code-smells/) (my new word of the week haha), but I'm thinking it couldn't hurt to have the Dockerfile checked in. Thoughts?

kittyflat (commented):
No longer have permissions to remove myself as reviewer

PR #229, opened 2020-01-19
Nora header tags typings

roblevintennis:
# Once corresopnding [Nora PR](https://github.com/getethos/ethos/pull/2021) approved cut another EDS version here, and then have the Nora PR point to that

**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1157190385219581/f)
- This is a part of the Nora Header work. We will need Tag and Header typings so we can consume in our Nora Typescript.
- Ran `yarn test:types` to verify

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.57",
+  "version": "1.1.58",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -15,6 +15,49 @@ export { Layout } from './Layout/index.js'
 // TODO -- DELETE THIS -- NOT GOING TO EXPORT DEPRECATED MEDIA
 // export { Media } from './Media'
 
+declare module 'react' {
+  interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
+    name?: string
+  }
+}
+
+interface HeaderProps {
+  name: string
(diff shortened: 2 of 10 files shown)


src/components/index.d.ts
```diff
@@ -15,6 +15,64 @@ export { Layout } from './Layout/index.js'
...
+  propTypes: {
+    /** a required unique name for the header */
+    name: string
+    /** Optional children to render on left side of header */
+    leftChildren: React.ReactNode
+    /** Optional children to render on right side of header */
+    rightChildren: React.ReactNode
+  }
+  defaultProps: {
+    leftChildren: any
```
roblevintennis: It's defined as `undefined` in the implementation. I'm honestly afraid to touch this as it was modified from running `tsc --jsx react --declaration` which is usually my first step in creating these type definitions. Also, we won't be using `defaultProps` when we consume this so it kind of doesn't matter.

Someday EDS will be fully in Typescript (or not?) lol

src/components/index.d.ts
```diff
@@ -245,7 +303,6 @@ export declare namespace ValueProps {
         })[]
   }
 }
-export default ValueProps
```
roblevintennis: This I believe was a mistake as you can only have one default export per file and this is a typings file so...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.57",
+  "version": "1.1.58",
```
stephaniel96: ah I didn't do this properly 😞

going to put up a PR to try and fix it. sorry @roblevintennis! n00b mistake

PR #231, opened 2020-01-21
Nora colors and Tag additions

stephaniel96:
**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1158041380837745)
- This PR adds Nora colors and also updates the `Tag` component to have all the colors we need. I decided to make the `Tag` component colors independent of status. In other words, the props we are passing are now colors e.g .`green` instead of `approved`.
- Tom and I worked on a color naming convention that would make sense for the Nora app. What we came up with is [here](https://www.figma.com/file/2Ulqfby7GNX8xU2yA2KeRW/%C2%B7%E2%80%80%E2%80%80%E2%80%80%E2%80%80%E2%80%80Nora%E2%80%80%E2%80%80%E2%80%80%E2%80%80%E2%80%80%C2%B7?node-id=790%3A122):
image

- 

**Screenshots:**

### DIFF
--- a/src/nora/components/Tag/Tag.js
+++ b/src/nora/components/Tag/Tag.js
@@ -5,11 +5,17 @@ import styles from './Tag.module.scss'
 export const Tag = ({ children, type }) => {
   let colorClass
   switch (type) {
-    case 'approved':
-      colorClass = styles.Approved
+    case 'red':
+      colorClass = styles.Red
       break
-    case 'rejected':
-      colorClass = styles.Rejected
+    case 'orange':
+      colorClass = styles.Orange
+      break
+    case 'green':
+      colorClass = styles.Green
+      break
+    case 'blue':
+      colorClass = styles.Blue
       break
     default:
       colorClass = styles.Default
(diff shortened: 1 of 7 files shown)


stephaniel96:
One thing to note is that I almost committed this code without updating _all_ usages of the old colors. Some sort of check (not sure if it exists already) to make sure that all variables exist would be great!

src/nora/components/Tag/Tag.js
```diff
@@ -5,11 +5,17 @@ import styles from './Tag.module.scss'
 export const Tag = ({ children, type }) => {
   let colorClass
   switch (type) {
-    case 'approved':
```
roblevintennis: The `propTypes` were not updated to reflect this so we get console warnings as approved and rejected are currently only valid inputs.

Easy peasy fix and I’m just starting my day so I’ll update Tag.js 🙂

I may add some test coverage too

stephaniel96: Got it! Thanks Rob :)

PR #232, opened 2020-01-21
Adding the Base Modal component, tons of a11y hooks

bwalker8038:
**Description:**

- [Asana Task](https://app.asana.com/0/1138309829121647/1150717201003937)

This `PR` adds as base modal component, as well as a bunch of `a11y` hooks

image

### Changes
- Added the `<Modal />` component and it's supporting files
- Added the `useTrapFocus` hook and supporting tests
- Added the `useBodyScrollLock` hook and it's supporting tests
- Added the `useSetFocus` hook and it's supporting tests
- Added the `useOutsideClick` hook and it's supporting tests
- Added the `useHideAriaSiblings` hook and it's supporting tests


src/hooks/useSetFocus.js
```diff
@@ -0,0 +1,34 @@
...
+ * element when the component is unmounted
+ *
+ * @public
+ *
+ * @param {React.MutableRefObject<HTMLElement | null>} ref - the coponent's ref
+ * @param {boolean} isFocusReady - if true, the ref can be focused
+ *
+ * @return {boolean} - a boolean that represents if the component is currently focused
+ */
+function useSetFocus(ref, isFocusReady) {
```
tkarssli: I don't see this Hook used anywhere, was that intentional?

src/hooks/a11y/useTrapFocus.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Handler for tab/shift+tab key presses and increments or
+   * decrements the `focusedElemIdx`.
+   *
+   * @param {number} lastIndex - the lastIndex that can be selected in the `elements`
+   * array
+   *
+   * @return {(e: KeyboardEvent) => void}
+   */
+  const handleTrapTab = (lastIndex) => {
```
tkarssli: Non-blocker but opening then clicking to establish focus in the modal and making the first key press a shift-tab focuses stuff in the background

src/components/Modal/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+
+Modal.propTypes = {
+  /** Id of an element that labels the modal */
+  ariaLabelledBy: PropTypes.string,
+  /** Id of an element that describes the modal */
+  ariaDescribedBy: PropTypes.string,
+  /** The Modal's children */
+  children: PropTypes.node.isRequired,
+  /** Boolean that sets the state of the modal */
+  showModal: PropTypes.bool,
```
tkarssli: Should this be required

src/components/Modal/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+  ariaDescribedBy,
+  ariaLabelledBy,
+  children,
+  toggle,
+  showModal,
+}) => {
+  const modalRef = useRef(null)
+
+  // Hooks! All the thingggs!!!!!
+  useOutsideClick(modalRef, () => toggle(false))
```
tkarssli: This could be an `onDismiss` or something since it only ever "toggles" off.

src/hooks/a11y/useHideAriaSiblings.js
```diff
@@ -0,0 +1,70 @@
...
+  const parent = elem.parentElement
+  if (parent === document.body) {
+    return elem
+  }
+
+  return getRootParent(parent)
+}
+
+/**
+ * Hook set's all nodes directly below the body, that are not not the `refs` parent tree,
```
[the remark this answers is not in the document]
wafflehorse: I'm gonna pretend that was intentional

src/hooks/a11y/useTrapFocus.js
bwalker8038: ah, good catch!, I'll take a look

src/hooks/useSetFocus.js
bwalker8038: That's correct, I wrote this hook for a previous working version of the `modal` and is not used in the current implementation. 

I included the hook in this PR since it is a reusable and useful hook when focus needs to be conditionally.

src/components/Modal/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+  useTrapFocus(modalRef, showModal)
+  useBodyScrollLock(showModal)
+  useHideAriaSiblings(modalRef, showModal)
+
+  return (
+    <div
+      role="dialog"
+      aria-modal="true"
+      aria-describedby={ariaDescribedBy && ariaDescribedBy}
```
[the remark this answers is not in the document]
bwalker8038: This should short circuit the prop assignment if `ariaDescribedBy` is `falsy`.

src/hooks/a11y/useHideAriaSiblings.js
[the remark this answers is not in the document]
bwalker8038: it was not

src/components/Modal/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+      className={styles.ModalContent}
+      ref={modalRef}
+    >
+      {children}
+    </div>
+  )
+}
+
+ModalContent.propTypes = {
+  ariaDescribedBy: PropTypes.string,
```
roblevintennis: Which of these are required vs. optional? I think `isRequired` my be needed for a couple of these

src/components/Modal/Modal.js
roblevintennis: he has a default parameter to `false`

src/components/Modal/Modal.js
```diff
@@ -0,0 +1,126 @@
...
+          </ModalContent>
+        )}
+      </div>
+    </Portal>
+  )
+}
+
+Modal.propTypes = {
+  /** Id of an element that labels the modal */
+  ariaLabelledBy: PropTypes.string,
```
roblevintennis: Should we have corresponding `defaultProps` for optional parameters (or just require the aria ones here?)

src/components/Modal/Modal.md
```diff
@@ -0,0 +1,79 @@
+## Description
+The `<Modal />` component provides a container that renders content into an accessible modal outlet. By default, the container will render `children` centered in the screen with an overlay. Out of the box, the `<Modal />` will handle setting proper focus on tabable elements within the inner content, trapping focus within the `modal`,  and toggle the state of the modal through keyboard events.
```
roblevintennis: what's a "modal outlet"?

src/components/Modal/Modal.md
```diff
@@ -0,0 +1,79 @@
...
+
+The container makes no assumptions about the inner content and can be styled to fit whatever implementation is needed.
+
+## Examples
+
+### Basic Usage
+Activating the modal is done through passing a `boolean` flag as the `showModal` `prop`. This gives the parent the ability to set if the modal will render open or closed on initial render.
+
+Toggling the state of the modal is also handled by passing a callback through the `toggle` `prop`. The callback is a `voided` function that accepts a  `boolean` as whether the modal is visible. When either the `escape` key is hit, or outside of the modal content is clicked, the modal will pass a boolean value of `false` to the `toggle` method.
```
roblevintennis: The part:

> is also handled by

confused me; I think `toggle` may be a slightly confusing name since there are two things happening (I think):

- Toggle the modal open or closed (action)
- Toggle the state (question e.g. is it open or closed?)

My understanding of the code is that:

- `showModal` will perform the action of showing/hiding the modal based on the boolean arg any time the Modal is rendered
- `toggle` is a callback that specifically receives a status boolean indicating whether the modal is visible or not

Is there a reason to pass in a callback other than a state hook setter? Should we name this more specifically to reduce confusion e.g. `toggleIsOpen` or something?

src/components/Modal/Modal.md
```diff
@@ -0,0 +1,79 @@
...
+      </>
+    )
+  }
+
+  <ModalExample />
+```
+
+### Setting `aria-describedby` and `aria-labelledby`
+
+Since the modal makes no assumptions about the content being rendered to the outlet, setting the `aria-describedby` and `aria-labelledby` properties are left up to the implementation.  The `<Modal />` does accept to optional properties, `props.ariaDescribedBy` and `props.ariaLabelledBy` which will set their respective `aria` properties on the inner modal to whatever `id` selector is passed to them.
```
roblevintennis: I'm still getting hung up on this word _outlet_. I googled and found [outlet](https://guides.emberjs.com/v1.10.0/cookbook/user_interface_and_interaction/using_modal_dialogs/) and I'm sure it's idiomatic, but I have not encountered this. Mind adding a definition?

src/components/Modal/Modal.md
```diff
@@ -0,0 +1,79 @@
...
+    const [isOpen, setIsOpen] = useState(false)
+
+    const HEADER_ID = 'long-modal-heading'
+    const DESC_ID = 'long-modal'
+
+    return (
+      <>
+        <button onClick={() => setIsOpen(true)}>Open Modal</button>
+        <Modal showModal={isOpen} toggle={setIsOpen} ariaLabelledBy={HEADER_ID} ariaDescribedBy={DESC_ID}>
+          <div style={{ background: '#fff', padding: '16px 24px', overflow: 'scroll', width: '100vw', height: '100vh' }}>
```
roblevintennis: You can import css into the styleguide (see `src/nora/components/Example/Example.module.scss` for example) and I'd prefer we don't encourage inline styling if possible

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  &Active {
+    align-items: center;
+    background: var(--GraySecondary--translucent);
+    display: flex;
+    height: 100vh;
+    justify-content: center;
+    left: 0;
+    position: fixed;
+    top: 0;
+    width: 100vw;
```
roblevintennis: Any reason not to put the height and width closer together here?

src/components/Modal/Modal.js
bwalker8038: `children`, `toggle`, and `showModal` are the only required props for `ModalContent`. I'm not sure how we want to handle these cases as `ModalContent` is an internal implementation of the `Modal` and not exposed as part of the public API

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+}
+
+.ModalContent {
+  position: relative;
+  overflow: scroll;
```
roblevintennis: What if we don't have enough content to scroll? Should we use `overflow: auto` here instead?

src/components/Modal/Modal.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const { getByTestId } = render(
+      <Modal {...props}>
+        <h1>Ima modal</h1>
+      </Modal>
+    )
+
+    const modal = getByTestId('base-modal-container')
+
+    act(() => {
+      fireEvent.keyDown(modal, { key: 'Escape', keyCode: 27 })
```
roblevintennis: First time I used testing library was because `fireEvent` "just works". Glad you just added it 👍

src/hooks/a11y/useBodyScrollLock.js
```diff
@@ -0,0 +1,30 @@
...
+  `overflow: hidden; position: fixed; width: 100%; top: ${-scrollPosition}px;`
+
+/**
+ * Hook set's scroll lock on the body element
+ *
+ * @param {boolean} setBodyOverflow - flag set's if the body is currently scroll locked
+ *
+ * @return {void}
+ */
+function useBodyScrollLock(setBodyOverflow) {
```
roblevintennis: I'm realizing I've done a horrible job of exposing our hooks and validators!!

Would you mind trying to add this to the styleguide.config.js?

```js
    {
      name: 'Hooks',
      content: './src/hooks/hooks.md',
      components: './src/hooks/**/*.js',
      sectionDepth: 2,
    },
    {
      name: 'Validators',
      content: './src/validators/validators.md',
      components: './src/validators/**/*.js',
      sectionDepth: 2,
    },
```

and just put placeholders for the markdown files?

src/components/Modal/Modal.js
bwalker8038: So the aria attributes are not required, but in most cases should be used. For instance, in a case where a form is included in the `modal`, `aria-describedBy` would not be needed for that case.

src/hooks/a11y/useBodyScrollLock.js
```diff
@@ -0,0 +1,30 @@
...
+    const scrollPos = window.scrollY
+
+    if (setBodyOverflow) {
+      body.setAttribute('style', LOCK_SCROLL_STYLES(scrollPos))
+    }
+
+    return () => {
+      const scrollY = body.style.top
+      body.setAttribute('style', '')
+      window.scrollTo(0, parseInt(scrollY || '0') * -1)
```
roblevintennis: It might be nice to have a short comment for this line

src/components/Modal/Modal.md
bwalker8038: ah, the `portal` that the modal content renders in to.

src/hooks/a11y/useHideAriaSiblings.js
```diff
@@ -0,0 +1,70 @@
...
+ * Returns the root node
+ *
+ * @private
+ *
+ * @param {HTMLElement} elem - the element with the parent to be checked
+ *
+ * @return {HTMLElement}
+ */
+function getRootParent(elem) {
+  const parent = elem.parentElement
```
roblevintennis: Nit: might be interesting to use iterative approach with same base case. I suppose there's no chance of overflow since there will always be a `document.body`, but a very deeply nested component might have a high space complexity I dunno

src/hooks/a11y/useHideAriaSiblings.js
roblevintennis: Awesome!

src/hooks/a11y/useTrapFocus.js
```diff
@@ -0,0 +1,80 @@
+import { useState, useEffect } from 'react'
+
+const FOCUSABLE_ELEMENTS_STRING =
+  'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, embed, [tabindex="0"], [contenteditable]'
```
roblevintennis: Nit: Where's you come up with this from? It's a bit scary in web world e.g. what if something were added or removed. Maybe at least just a link back to source would be worthwhile.

roblevintennis: This matrix is kinda interesting for this:

https://allyjs.io/data-tables/focusable.html

src/hooks/a11y/useTrapFocus.js
```diff
@@ -0,0 +1,80 @@
...
+      if (e.key === 'Tab' || e.keyCode === 9) {
+        e.preventDefault()
+
+        // if the index is the last elem of the focusableElems, focus on the first elem,
+        // otherwise, increment the index
+        let idx = focusedElemIdx === lastIndex ? 0 : focusedElemIdx + 1
+
+        // if the index is 0 and the shift+tab is hit – focus on the last elem,
+        // otherwise, decrement the index
+        if (e.shiftKey) {
```
roblevintennis: since you've been using the pattern of falling back on keyCode is that required here or is `shiftKey` the same across envs?

src/hooks/useSetFocus.js
roblevintennis: Yeah, good catch. Although I really like this hook, it might be best to wait for a use case 🤔 unless you can find a place in the codebase and redirect it to use this 😋 

I dunno, I'll just leave it up to Brad. Be nice to get feedback on the feedback though :)

src/hooks/a11y/useTrapFocus.js
bwalker8038: This is one of the few cases where a keyboard event is supported across the board: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey

roblevintennis: ahhh, good guess, huh!? 👍

src/components/Modal/Modal.md
bwalker8038: 👍

roblevintennis:
Also, I feel like somehow maybe the "global-ness" of the hooks specs might have caused other stuff to fail? Not sure why CI/CD travis is upset there

src/components/Modal/Modal.md
bwalker8038: I'm not sure what a good name for this pattern would be in `React` 🙃 . `outlet` is definitely a Ember-ism. I guess I could rename it to `Modal Portal`? 🤷‍♂

src/components/Modal/Modal.js
bwalker8038: I agree with that –  looking at the state of the component now, we the modal does only set the toggle state to `false`.

src/components/Modal/Modal.md
bwalker8038: I agree, as Tamir pointed out above, internally, the modal only ever passes `false` to the `toggle` method. It would make sense to name this `prop` something like `onDismiss`. `showModal` could be renamed to `isActive` or `isShown`

There might be times where you might want to pass a function to `toggle` that's not just a state setting, e.g. maybe when you dismiss the modal, you want to fire some sort of tracking event.

src/components/Modal/Modal.module.scss
bwalker8038: The properties are listed in alphabetical order. There's really no reason for this other than it being a habit I picked up years ago...

src/components/Modal/Modal.module.scss
bwalker8038: Yep, it should be. The interesting thing is I didn't catch this because macOS breaks this behavior by default

src/components/Modal/Modal.md
roblevintennis: Yeah, on second read `onDismiss` sounds like a winner 👍

src/components/Modal/Modal.md
roblevintennis: So you could just update this with:

`...into an accessible modal outlet (the portal that the modal content renders in to)`

src/components/Modal/Modal.js
roblevintennis: Ah gotcha, then maybe just remove the PropTypes for the internal ModalContent

src/components/Modal/Modal.js
roblevintennis: Yeah, disregard my other comment I like `onDismiss` :)

src/hooks/a11y/useHideAriaSiblings.js
bwalker8038: I might be wrong – but I think in most use cases, the recursive calls should not be an issue. The  tree would need to be fairly deep to cause an overflow, and the use case of this hook would be on a component like the modal that's fairly close to the `body`.

Funny enough – this would only be an issue in ever engine but Webkit. Safari implements TCO in safari 13

src/hooks/a11y/useTrapFocus.js
bwalker8038: added a comment

src/components/Modal/Modal.module.scss
tkarssli: Oh nice never seen that. I can see how that is useful especially with really long styles

src/components/Modal/Modal.js
bwalker8038: I think it's fine to leave them for now, I'd rather not disable the eslint rule for that block. This should be a mute point when we move completely to typescript 🙃

src/components/Modal/Modal.module.scss
roblevintennis: that's fine...it's just a nitpick / non-blocker

PR #233, opened 2020-01-21
Document EDS

roblevintennis:
Why: EDS is sorely lacking documentation for contributing, how the form engine works, etc.
How: Adds a Guidelines section with contribute and a11y pages. a11y is TBD
Tags: documentation, guidelines

**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1158167717957553/f)
- EDS Docs


README.md
```diff
@@ -49,6 +43,32 @@ Then do following to setup debugging in Chrome:
...
+Run only tests related to changed files based on git:
+
+```shell
+yarn test -o
+```
+
+Find tests related to `fileA.js`:
+
+```shell
+jest --findRelatedTests path/to/fileA.js
```
kittyflat: i think this is also `yarn test`

README.md
```diff
@@ -109,12 +129,11 @@ the server where Jenkins lives will cache these and attempt to reuse them if it
 
 #### CSS/SCSS concepts
```
roblevintennis: Nick wrote `CSS/SCSS concepts` and I've always been a bit on the fence about it 😏 ... my changes here are just formatting

src/guidelines/a11y.md
```diff
@@ -0,0 +1 @@
+TBD
```
roblevintennis: @bwalker8038 I've got some stuff for this I can add but definitely would like you to piggyback with your expertise sir!

src/guidelines/contribute.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Exposing Components
+
+In order for EDS components to be consumable via something like `import { MyComponent } from 'ethos-design-system'`, `MyComponent` will need to have been added to the component exports [index.js](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.js) file.
+
+### Typescript Definitions
+
+If your consumer app is using Typescript, it will require that the component you wish to use has a corresponding `index.d.ts` Typescript [definition](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts)—we will discuss this more in [How to submit your contribution](#how-to-submit-your-contribution) section that follows.
+
+### Form engine components
```
roblevintennis: @wafflehorse I did my best to at least provide some pointers to important concepts, files, and basically give an overview on all this. It turns out to be challenging to document without getting very long-winded. Would this section have helped you a couple weeks back and/or do you see an egregious omission?

Keep in mind these docs will hopefully evolve over time iteratively

README.md
roblevintennis: really? I have to admit I don't use `--findRelatedTests` tests much but do `-t` or `yarn test -t REGEX` which does similar stuff

src/guidelines/contribute.md
wafflehorse: I think you nailed all the critical points. This would have definitely helped with the Select component. Thanks for adding on this!

PR #234, opened 2020-01-21
Loader II

zach-ethos:
Stragglers from the last PR.


src/components/index.js
```diff
@@ -13,6 +13,7 @@ export { Form } from './Form'
 export { Grid, Pagination, Row, Column, useGridSorting } from './Grid'
 export { InfoMessage } from './InfoMessage'
 export { Layout } from './Layout'
+export { Loader } from './Loaders/Loader'
```
roblevintennis: Awesome, thanks Zach! So are we consuming this directly as typescript? I somehow thought `src/components` might choke if we added typescript directly, but maybe not?

zach-ethos: It might! Let's chat IRL

PR #236, opened 2020-01-22
Add setSortState function for arrow control

stephaniel96:
**Description:**
- [Ticket](https://app.asana.com/0/1150068311310825/1158253271849323)
- We aren't using the `compareBy` function when using the Grid in Nora because we have to re-fetch the page rather than sorting locally. 
- As a result, we need a way to set the sort state so that the correct sort icons appear.
- This PR adds the ability to set the sort state from outside of the EDS component.

**Screenshots:**
- Tested locally to make sure that the sort still works and there are no console errors.
Image 2020-01-22 at 12 01 13 PM

- https://github.com/getethos/ethos/pull/2046

### DIFF
--- a/src/components/Grid/useGridSorting.js
+++ b/src/components/Grid/useGridSorting.js
@@ -26,6 +26,35 @@ export const useGridSorting = (rows, columns) => {
     return <SortIcon columnKey={key} />
   }
 
+  const setSortState = (key) => {
+    let nextSort
+    let currSortCopy = currentSort
+
+    // Have we sorted by this key before? If so, follow next state rules.
+    if (currSortCopy[key]) {
+      if (currSortCopy[key].direction === 'down') {
+        nextSort = 'up'
+      } else if (currSortCopy[key].direction === 'up') {
+        nextSort = 'down'
+      }
+      currSortCopy[key].direction = nextSort
+    } else {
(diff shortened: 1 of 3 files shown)


src/components/Grid/useGridSorting.js
```diff
@@ -26,6 +26,35 @@ export const useGridSorting = (rows, columns) => {
     return <SortIcon columnKey={key} />
   }
 
+  const setSortState = (key) => {
```
stephaniel96: @roblevintennis couldn't figure out how to write specs mostly because the original logic of `compareBy` didn't change and when trying to write tests for `setSortState` I was beginning to test implementation lol. Let me know what you think!

roblevintennis: Yeah, I had a look and I don't think we can easily get coverage for `setSortState` since `testHook` which creates a parent component so we can abide by React Hooks rules of only using a hook within a top level component will rerender in an infinite loop (unsurprisingly). I think other libs might start to make that easier and Brad has an open PR which might be adding one of them. I think it's fine for now....

But, we can certainly get _some_ coverage :) I'm cooking up a PR https://github.com/getethos/ethos-design-system/pull/238 and I'll try pulling that into your list refactor 4 branch and ensure I haven't borked anything.

PR #245, opened 2020-01-24
Add support for fontawesome for github actions

alexthecarp:
Change github actions to set up .npmrc with fontawesome token.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,12 +10,6 @@ jobs:
     name: Build and Deploy
     runs-on: ubuntu-latest
     steps:
-      - name: Setup fontawesome dependencies
-          run: |
-            cat > .npmrc <<EOF
-            @fortawesome:registry=https://npm.fontawesome.com/
-            //npm.fontawesome.com/:_authToken=${{ secrets.FONTAWESOME_TOKEN }}
-            EOF
       - uses: actions/checkout@v1
       - name: Cache dependencies
         id: cache-dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -2,14 +2,20 @@ name: Deploy Styleguide
 
 on:
   push:
-    branches:
-      - master
+#    branches:
```
roblevintennis: Can we prune these comments?

.github/workflows/deploy.yml
```diff
@@ -2,14 +2,20 @@ name: Deploy Styleguide
...
+#      - master
 
 jobs:
   build:
     name: Build and Deploy
     runs-on: ubuntu-latest
     steps:
+      - name: Setup fontawesome dependencies
+          run: |
```
roblevintennis: I'm not sure why it's upset: `You have an error in your yaml syntax on line 14`

.github/workflows/deploy.yml
```diff
@@ -24,6 +18,12 @@ jobs:
           path: node_modules
           # Don't share with other workflows because we are using `--production=false`
           key: dev-dependencies-${{ hashFiles('**/yarn.lock') }}
+      - name: Setup fontawesome dependencies
```
roblevintennis: Does moving it here allow the node modules to stay cached thus speeding things up I'm guessing?

.github/workflows/deploy.yml
roblevintennis: `You have an error in your yaml syntax on line 22` I'm guessing it's a tabs / spaces nuanced issue

.github/workflows/deploy.yml
alexthecarp: no, they are here for a reason. normally we want the branch protection, but for any change of this yml file, if i leave that in then it wont' run until i merge. SO I have to comment them out, get it working and finally, as a last step, I uncomment these before checking in,

.github/workflows/deploy.yml
alexthecarp: uyeah.. ugh

.github/workflows/deploy.yml
alexthecarp: Yeah it feels like a better place. Though I'm not sure.

PR #254, opened 2020-02-04
Nav Bar covers top part of modal

iamvictorli:
**Description:**
Navbar covers top part of modal

Current interaction with Navbar and Modal
Screen Shot 2020-02-03 at 5 53 14 PM

Fixed
Screen Shot 2020-02-03 at 5 53 31 PM

- Using EDS Modal on CMS

**Manual Testing**
- Navbar and Modal interaction
- Universal Navbar still looks fine

**Screenshots:**

### DIFF
--- a/src/components/UniversalNavbar/UniversalNavbar.js
+++ b/src/components/UniversalNavbar/UniversalNavbar.js
@@ -100,95 +100,84 @@ class UniversalNavbar extends React.Component {
     )
 
     return (
-      <div style={{ height: 64 }}>
-        <div
-          style={{
-            position: 'fixed',
-            top: 0,
-            left: 0,
-            width: '100%',
-            height: 100,
-            zIndex: 1,
-          }}
-        >
-          <Layout.ScrollDetector>
-            <div className={styles.hamburger}>
-              <TransformingBurgerButton
-                showMobileMenu={showMobileMenu}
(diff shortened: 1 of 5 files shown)


src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -100,95 +100,84 @@ class UniversalNavbar extends React.Component {
...
-            zIndex: 1,
-          }}
-        >
-          <Layout.ScrollDetector>
-            <div className={styles.hamburger}>
-              <TransformingBurgerButton
-                showMobileMenu={showMobileMenu}
-                clickHandler={this.toggleHamburger}
-              />
+      <div className={styles.navbar}>
```
iamvictorli: Moved to CSS Modules

src/components/UniversalNavbar/UniversalNavbar.js
```diff
@@ -100,95 +100,84 @@ class UniversalNavbar extends React.Component {
     )
 
     return (
-      <div style={{ height: 64 }}>
```
c7hudson: @iamvictorli any idea why we had this extra div?

iamvictorli: No I don't. I don't think this is needed as there is another div that has a height.

PR #265, opened 2020-02-21
EDS: Nora Button

stephaniel96:
**Description:**
- [Asana Task](https://app.asana.com/0/1150068311310825/1162925575699393)
- Creates a default EDS Nora Button that allows for customization (given that lengths margin/padding etc may differ)
- NOTE: DO NOT MERGE until it is approved AND a tag is added to the PR. 

image

-

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -366,7 +366,8 @@ export declare namespace ValueProps {
           header: string
           subHeader: string
           alt?: undefined
-        })[]
+        }
+    )[]
   }
 }
 
@@ -890,15 +891,39 @@ export declare const Header: {
   }
 }
 
+type ButtonType = 'button' | 'submit'
+
 interface NoraButtonProps {
-  className: string
-  children: React.ReactNode
+  children?: string
+  'data-tid'?: string
+  disabled?: boolean
+  type?: ButtonType
+  name?: string
+  onClick?: any
+  role?: string
+  className?: string
 }
+
 export declare const NoraButton: {
-  ({className, children}: NoraButtonProps): JSX.Element
+  ({
+    children,
+    disabled,
+    type,
+    name,
+    onClick,
+    role,
+    className,
+    ...rest
+  }: NoraButtonProps): JSX.Element
   propTypes: {
-    className: string
     children: React.ReactNode
+    'data-tid': string
+    disabled: boolean
+    type: ButtonType
+    name: string
+    onClick: any
+    role: string
+    className: string
   }
 }
 
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -890,6 +890,18 @@ export declare const Header: {
   }
 }
 
+interface NoraButtonProps {
+  className: string
+  children: React.ReactNode
+}
+export declare const NoraButton: {
+  ({className, children}: NoraButtonProps): JSX.Element
+  propTypes: {
+    className: string
+    children: React.ReactNode
+  }
+}
+
 interface TagProps {
   type: string
   children: React.ReactNode
--- a/src/nora/components/NoraButton/NoraButton.js
+++ b/src/nora/components/NoraButton/NoraButton.js
@@ -2,13 +2,39 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styles from './NoraButton.module.scss'
 
-export const NoraButton = ({ className, children }) => {
+export const NoraButton = ({
+  className,
+  disabled,
+  type,
+  name,
+  onClick,
+  role,
+  children,
+  ...rest
+}) => {
+  const cssClasses = [styles.Button, className].join(' ')
   return (
-    <button className={[styles.Button, className].join(' ')}>{children}</button>
+    <button
+      className={cssClasses}
+      disabled={disabled}
+      type={type}
+      name={name}
+      onClick={onClick}
+      data-tid={rest['data-tid']}
+      role={role}
+    >
+      {children}
+    </button>
   )
 }
 
 NoraButton.propTypes = {
+  children: PropTypes.string,
+  'data-tid': PropTypes.string,
+  disabled: PropTypes.bool,
+  type: PropTypes.oneOf(['button', 'submit']), // See Button.js for explanation of these two types
+  name: PropTypes.string,
+  onClick: PropTypes.func,
+  role: PropTypes.string,
   className: PropTypes.string,
-  children: PropTypes.string.isRequired,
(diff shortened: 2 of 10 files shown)


src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  line-height: var(--Space-32);
+  letter-spacing: 0.015em;
+  color: white;
+  background: var(--NoraBlue-400);
+  border-radius: 3px;
+  text-align: center;
+  display: inline-block;
+}
+
+/* Used only for md example */
```
roblevintennis: You can create an `example.scss` file and pull it in via es6 import right from the .md example. We've opted for that to keep example css out of the library ones

src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
+.Button {
+  font-weight: 500;
+  font-size: 13px;
```
roblevintennis: We have a compact font that's been used for Nora so far and it's 15 px. If we want to add a 13px (I presume you verified some figma designs or something), then we should figure out a way to do that in a sharable way across other components that'll need it so we have a sort of system if you will. Maybe it's  18px | 15px | 13px or whatever. But I'd prefer we centralize it somehow.

src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  height: var(--Space-32);
+  line-height: var(--Space-32);
+  letter-spacing: 0.015em;
```
roblevintennis: I've found letter spacing css to be so arbitrary, I've pretty much given up on kerning and tracking on the web unless it's a display heading or something that needs to be curated.

src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  color: white;
```
roblevintennis: we have a `--White` I believe

src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  background: var(--NoraBlue-400);
+  border-radius: 3px;
```
roblevintennis: We have a var for this (I thought you put it in Attributes when you did the tags stuff)

src/components/index.d.ts
```diff
@@ -890,6 +890,18 @@ export declare const Header: {
   }
 }
 
+interface NoraButtonProps {
+  className: string
+  children: React.ReactNode
+}
+export declare const NoraButton: {
```
roblevintennis: I'm pretty sure we'll need to expose some events on these buttons at minimum `onClick`

src/nora/components/NoraButton/NoraButton.module.scss
stephaniel96: Hm, good to know. I just verified that the mocks have 13px for buttons (it seems like when things are bolded the font sometimes is smaller), so perhaps we can bring this up with Tom.

Would you prefer for me to change it to 15 px for now?

src/nora/components/NoraButton/NoraButton.module.scss
stephaniel96: Oh cool! Didn't see that

src/components/index.d.ts
stephaniel96: oh true 🤦‍♀ will do that now

src/nora/components/NoraButton/NoraButton.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from './NoraButton.module.scss'
+
+export const NoraButton = ({ className, children }) => {
+  return <button className={[styles.Button, className].join(' ')}>{children}</button>
```
roblevintennis: Can you look into adding the following props on the button?

```
      disabled={disabled}
      type={type}
      name={name}
      onClick={onClick}
      data-tid={rest['data-tid']}
      role={role}
```

Most of these should be self-evident, but `data-tid` is specifically for unit and cypress

stephaniel96: will do! don't know what I was thinking with this earlier smh

src/nora/components/NoraButton/NoraButton.module.scss
roblevintennis: No, I think 13px is probably correct. I'd just like to have it defined so we can do something like `font-size: var(--x-compressed-font-size) * 1px` or something  similar 🤔It's a bit hard...we already conveniently had the compact one which I repurposed for Nora's main font size, but, we'll need to add one for the 13px variant.

src/nora/components/NoraButton/NoraButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  text-align: center;
+  display: inline-block;
+}
```
roblevintennis: On my branch for the Notes textarea that this will soon replace, I'm doing:

```scss
.Submit[disabled] {
  opacity: 0.65;
}
```

so you should probably do here:

```
.Button[disabled] {
  opacity: 0.65;
}
```

I copied `.65` from bootstrap but I could be persuaded to use `.7` if we prefer :)

stephaniel96: hm, should these styles be passed in through the className instead? or does that not work

src/nora/components/NoraButton/NoraButton.module.scss
stephaniel96: I'll add a 13px font attribute to the `Attributes.scss` file that we have for Nora, and perhaps as a follow-up action, we can take a holistic look at this with Tom?

src/components/index.d.ts
```diff
@@ -366,7 +366,8 @@ export declare namespace ValueProps {
           header: string
           subHeader: string
           alt?: undefined
-        })[]
```
stephaniel96: not sure why my prettier linted this as well 😬

roblevintennis: Doesn't bother me I dont think

stephaniel96:
> > I'll add a 13px font attribute to the Attributes.scss
> 
> On this PR? I think it'd be preferable if you don't mind.

Done!

PR #266, opened 2020-02-24
Nora button and icon link followup

roblevintennis:
**Description:**
- After merging 265 I realized we had some things missing from there in terms of exports and what not. Also, when we changed IconLink prop, we never updated the type defs accordingly

- https://github.com/getethos/ethos-design-system/pull/265

**Screenshots:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.95",
+  "version": "1.1.96",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.96",
+  "version": "1.1.97",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.93",
+  "version": "1.1.94",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.94",
+  "version": "1.1.95",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -51,6 +51,7 @@ export { Modal } from './Modal'
 ////////////////////////////
 // ---- NORA EXPORTS ---- //
 ////////////////////////////
-export { IconLink } from '../nora/components/index'
 export { Header } from '../nora/components/index'
+export { IconLink } from '../nora/components/index'
+export { NoraButton } from '../nora/components/index'
 export { Tag } from '../nora/components/index'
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.93",
+  "version": "1.1.97",
```
gbockus: So if you move here to bump the version in the PR then create the tag off master after merge?

src/components/index.js
```diff
@@ -51,6 +51,7 @@ export { Modal } from './Modal'
 ////////////////////////////
 // ---- NORA EXPORTS ---- //
 ////////////////////////////
-export { IconLink } from '../nora/components/index'
 export { Header } from '../nora/components/index'
+export { IconLink } from '../nora/components/index'
+export { NoraButton } from '../nora/components/index'
```
gbockus: We have things named specifically for Nora in the EDS?   That seems like an odd direction.

package.json
roblevintennis: We err on the side of continuing to bump if we realize we need additional commits because yarn cache and Jenkins can get quite screwy otherwise. We can make a tag after the fact e.g. if you merge a PR, and then realize you forgot to tag it. Really, the only key is that there's no commits between the desired commit and the tag.

Not sure I understood your question or answered it :)

src/components/index.js
roblevintennis: We have a whole sub-section for these Nora based components. The idea is that Design wants the main product sort of "locked up", and these Nora component are likely not appropriate for say the main funnel app. It's sort of best we have now sort of thing.

gbockus: So why even put them in the EDS?

package.json
gbockus: mostly just making sure I understand when I need to do it.  I think we're run into the problem a few times in the past there was some confusion about when to create the tag.  It might be useful to have a script to bump the version and create a tag in the master branch just to make it stupid easy.

src/nora/components/NoraButton/index.js
```diff
@@ -0,0 +1 @@
+export { NoraButton } from './NoraButton'
```
stephaniel96: Can we write a hook or something that checks that these components are properly exported, or add this to a README somewhere?

src/nora/components/index.js
```diff
@@ -1,3 +1,4 @@
-export { IconLink } from './IconLink'
 export { Header } from './Header'
+export { IconLink } from './IconLink'
+export { NoraButton } from './NoraButton'
```
stephaniel96: Could we add this step to the README as well?

src/components/index.tests.tsx
```diff
@@ -26,8 +33,8 @@ class IconLinkTest extends React.Component<any, any> {
         iconName="exchange"
         iconClassName="iconClassName"
         iconContainerClassName="iconContainerClassName"
-        linkClassName="linkClassName"
-        linkPosition="right"
+        textClassName="linkClassName"
```
stephaniel96: @roblevintennis these tests weren't failing when I committed the PR updating these fields, and running `yarn test` seemed to pass.

If these tests aren't running and preventing commits, what is the point of them? I'm finding that there's a fair bit of cognitive overhead in order to add something to EDS, and without a checklist it's becoming near-impossible for me to remember all the things to do. 😅 Not sure if this is just me, but thought I'd surface in case there are others that feel this way!

stephaniel96: I guess the other thing is that it's hard for both the author and the reviewer to remember all these things haha. I'm also curious how you even found this issue? I can't get an error to throw whatsoever.

roblevintennis: These tests are not yet automated and just confirm the typings are correct. But we could certainly benefit in having them automated. Limited bandwidth and resources on EDS is the culprit. That said, if these are things you'd like to submit a PR for it would be awesome.

stephaniel96: Awesome and thanks for the context! Put up #268 :)

PR #270, opened 2020-02-27
Make labelCopy optional in ButtonSelectGroup

zach-ethos:


roblevintennis:
I think we’re inconsistent with this and usually DO require it. My sense is the button groups often don’t use labels so it’s an exception to the rule. Correct?

PR #272, opened 2020-02-28
BaseTextAreaInput and NoraTextAreaInput

stephaniel96:
**Description:**

- Asana Task: https://app.asana.com/0/1150068311310825/1164225131436481
- Create a `BaseTextAreaInput` component that is shared between `NoraTextAreaInput` and `TextAreaInput` components. 

**Is this a new component? Please review these reminders: **

_Please pick one and delete options that are not relevant:_

- [x] Have you read the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute)?
- [x] Have exported the component from the main [index.js](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.js)?
- [x] Have you exported it from the [index.d.ts](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts) so Typescript users can consume it? Added a test and ran `yarn test:types`?
- [x] Followed the checklist at the bottom of the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute) guide?

**Referencing PR or Branch (e.g. in CMS or monorepo):**

-

**Screenshots:**
image

PR #273, opened 2020-02-29
Make labels optional, pt. II

zach-ethos:
Recent designs don't include labels, and the fields shouldn't require or assume them. 

Also, I think labelCopy in defaultProps is an antipattern now, because you can't omit it now or even pass an empty string. We'd have to add another prop like `showLabel` and that seems unnecessary.

The only place I see this used in production is on the CMS and the partner flow. I am making PRs to put those folks in the loop.

### DIFF
--- a/src/components/TextMaskedInput/TextMaskedInput.js
+++ b/src/components/TextMaskedInput/TextMaskedInput.js
@@ -135,7 +135,9 @@ export const TextMaskedInput = (props) => {
 
   return (
     <>
-      <InputLabel name={name} labelCopy={labelCopy} allCaps={allCaps} />
+      {labelCopy && (
+        <InputLabel name={name} labelCopy={labelCopy} allCaps={allCaps} />
+      )}
       {getMaskedInputByType(mask)}
       {!doValidation && getError(currentError, whichTouched)}
     </>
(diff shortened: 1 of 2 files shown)


src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -135,7 +135,9 @@ export const TextMaskedInput = (props) => {
 
   return (
     <>
-      <InputLabel name={name} labelCopy={labelCopy} allCaps={allCaps} />
+      {labelCopy && (
```
roblevintennis: Should we go through all the fields and do this then? @stephaniel96 is about to jump through several hoops (if she didn't already) to get around this for Nora as well for the TextAreaInput. We also needed to expose some other stuff too specific for Nora so wrapping it in a `NoraTextAreaInput` might make sense anyway, but, I think we'll need to make labelCopy optional for all fields. Can you do an audit and advise on a game plan for follow ups and/or tickets?

jameshawkinsjr: I'm actually doing the same in one of my PRs -- I'm passing in `labelCopy=''` into a few components (specifically `BirthdateInput`,`ZipInput`, `TextInput`, but I'm still adding more)

stephaniel96: My current WIP is #272 and would be more than happy to make the label optional for the `TextAreaInput` component

roblevintennis:
@zach-ethos how come you closed this? I liked the direction :) Was it due to untimely review? If so, my bad and I can help push it through. I think we should definitely shoot for making `labelCopy` optional iteratively

PR #276, opened 2020-03-03
Accordion

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1162798638929451/f)
- EDS Accordian

- [x] Have you exported it from the [index.d.ts](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts) so Typescript users can consume it? Added a test and ran `yarn test:types`


src/components/Accordion/AccordionSection.js
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <>
+      <div
+        role="button"
+        aria-expanded={expanded}
+        aria-controls={sectionId}
+        id={labelId}
+        tabIndex={0}
+        className={styles.Label}
```
stephaniel96: I'm thinking about what needs we have for `Nora` and was wondering if we could make the style of the `title` customizable? Right now, Nora section headers have pretty particular styling

image

src/components/Accordion/AccordionSection.js
```diff
@@ -0,0 +1,102 @@
...
+        onClick={onClick}
+        onKeyDown={onKeyDown}
+        onFocus={onFocus}
+        onBlur={onBlur}
+        ref={labelRef}
+      >
+        {title}
```
stephaniel96: I'm thinking about what needs we have for `Nora` and was wondering if we could make the style of the `title` customizable? Right now, Nora section headers have pretty particular styling

image

roblevintennis: so you just want to be able to pass in a `className` right?

stephaniel96: Yeah I think that would work! I also wonder if we should separate `title` and `renderToggle(expanded)`. In the mocks right now, it seems to me that the click target is only the "+Add" text and not the entire row. Not sure what the ideal behavior should be though

roblevintennis:
We now have a second example in the Accordion page that shows how to use `toggleChildIsTarget` boolean prop to indicate that we should only toggle sections off the toggle (not the entire header bar):

image

image

src/components/index.tests.tsx
```diff
@@ -324,10 +339,6 @@ class ZipInputTest extends React.Component<any, any> {
   }
 }
 
-///////////////////////////////
```
roblevintennis: We're not keeping this ordering so I figured it'd be better to just remove this as it's not become misleading

src/components/index.d.ts
```diff
@@ -1037,6 +1037,39 @@ export declare const NoraDrawer: {
...
+  ({
+    index,
+    title,
+    renderToggle,
+    children,
+  }: {
+    index?: number
+    title: string
+    renderToggle?: any
+    children?: React.ReactNode
```
stephaniel96: In the component file, `children` seems to be a required `prop`.  I think it makes sense for `children` to be required?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,6 +11,7 @@ _Please pick one and delete options that are not relevant:_
 - [ ] Have exported the component from the main [index.js](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.js)?
 - [ ] Have you exported it from the [index.d.ts](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts) so Typescript users can consume it? Added a test and ran `yarn test:types`?
 - [ ] Followed the checklist at the bottom of the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute) guide?
+- [ ] Bumped the yarn version?
```
stephaniel96: Thanks!

src/components/index.d.ts
```diff
@@ -1037,6 +1037,39 @@ export declare const NoraDrawer: {
...
+  ({
+    id,
+    expanded,
+    onToggle,
+    children,
+  }: {
+    id?: string
+    expanded: object
+    onToggle: any
+    children: React.ReactNode
```
stephaniel96: I think the `Accordion` props need to be updated with the new `toggleChildIsTarget` prop

src/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,100 @@
...
+  onToggle: PropTypes.func.isRequired,
+  /** The Accordion's children. Likely AccordionSection's */
+  children: PropTypes.node.isRequired,
+  /** Boolean prop that indicates that you'd like to use that as the toggle (not the entire
+   * `AccordionSection` header bar). Defaults to `false` */
+  toggleChildIsTarget: PropTypes.bool,
+}
+
+Accordion.defaultProps = {
+  expanded: null,
```
stephaniel96: if `toggleChildIsTarget` defaults to `false`, should we have it in the `defaultProps` here?

src/components/index.d.ts
roblevintennis: 🤦‍♂ yup. fixing, thanks Steph

src/components/Accordion/Accordion.js
roblevintennis: yeah, or as a default prop in props but I'll just add here 👍

src/components/Accordion/useAccordionState.js
```diff
@@ -0,0 +1,12 @@
+import { useState } from 'react'
+
+export const useAccordionState = (initialState) => {
```
stephaniel96: Just to make sure I'm understanding everything, is `useAccordionState` only meant to be used in the example to show how `onToggle` and `expanded` props work?

src/components/Accordion/useAccordionContext.js
```diff
@@ -0,0 +1,11 @@
...
+export const AccordionContext = createContext({
+  focusRef: {},
+  selected: [null],
+  toggleChildIsTarget: false,
+  expandedAll: [],
+  onToggle: undefined,
+  onNavigation: () => undefined,
+  id: null,
+})
+export const useAccordionContext = () => useContext(AccordionContext)
```
stephaniel96: Most of the documentation on context that I was reading (https://reactjs.org/docs/context.html) described context before hooks were introduced and I was briefly confused. 😂 

This looks great to me!

src/components/Accordion/useAccordionState.js
roblevintennis: yeah, the consumer can set everything up with something like:
`const { expanded, onToggle } = useAccordionState({ 1: true })`

PR #282, opened 2020-03-04
Remove extra prop in NoraDrawer type definition

stephaniel96:
**Description:**

- Noticed this when consuming the `NoraDrawer` in the main app. 
- This is a simple fix and does NOT require a new release.

**Screenshots:**

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -1020,7 +1020,6 @@ export declare const NoraDrawer: {
     onDismiss,
     isOpen,
     position,
-    className,
     labelCopy,
     closeCopy,
   }: {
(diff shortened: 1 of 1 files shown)


src/components/index.d.ts
```diff
@@ -1030,7 +1029,6 @@ export declare const NoraDrawer: {
     labelCopy: string
     closeCopy: string
     position?: string
-    className?: string
```
stephaniel96: NoraDrawer does not have this prop. 

https://github.com/getethos/ethos-design-system/blob/0fb56f9e2680be162c6c11b8b4372d28126c4cfe/src/nora/components/NoraDrawer/NoraDrawer.js#L44-L50

PR #284, opened 2020-03-05
Snackbar component

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1163005231692857/f)
- EDS Snackbar component


src/components/Snackbar/Snackbar.module.scss
```diff
@@ -0,0 +1,47 @@
+@import '../Animation.scss';
```
roblevintennis: For the Nora component, we'll basically reimplement this to skin as Tom's designed things

src/components/Animatable.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.ZoomInUp {
+  animation-name: zoomInUp;
+}
+
+@keyframes zoomInUp {
+  from {
+    opacity: 0;
+    transform: scale3d(0.5, 0.5, 0.5) translate3d(0, 1000px, 0);
+    animation-timing-function: cubic-bezier(0.55, 0.055, 0.675, 0.19);
```
roblevintennis: These are pretty specific at the moment, and so I've elected to not share them yet in Animation.scss. If I start to see a repeatable pattern I will, but for now this is one-off

roblevintennis:
In working on this in tandem with Nora, I discovered that since Nora uses Nest.js, we have to account for both the Fetch API Response interface, and also the Nest.js one—Nest.js uses `statusCode` not `status` for example.

Here's manually testing after changes made, and the XHR examples all appear to still work properly:

image

src/nora/components/NoraTextAreaInput/NoraTextAreaInputExample.module.scss
```diff
@@ -9,7 +9,7 @@
 
 .TextClassExample {
   margin-top: var(--Space-12);
-  font-size: calc(var(--compact-font-size)* 1px);
+  font-size: var(--Nora-Font-15);
```
stephaniel96: 👍

PR #279, opened 2020-03-03
UniversalNavbarExpanded - fast follow cleanup

c7hudson:
**Todo before merge**
- [x] Merge master into feature branch and then bump version!
- [ ] QA on CMS 

**Description**
- Follows #274 to improve code quality, address original PR comments and design team QA tickets.
- [Asana Task](https://app.asana.com/0/1116481661798430/1158569544825133/f) | note sub-tasks prefixed with `QA`

**Disclaimer**
- I acknowledge the large diff and have attempted to document the changes as comprehensively as possible. I've also been actively QAing on the [CMS on this deploy preview](https://deploy-preview-2819--ethos-cms.netlify.com/). Here is a breakdown for why the diff is so large:
  - Breaking the main `UniversalNavbarExpanded.js` into multiple files contributed ~1100 newlines, although much of this is re-used code.
  - ~1400 total changes from `UniversalNavbarExpanded.test.js.snap`
  - 500 changes to the main `UniversalNavbarExpanded.js` after being decomposed.
  - ~300 changes to `UniversalNavbarExpanded.module.scss`, mostly migrating styles to individual `.module.scss` files.
- I'm going to try and pair with an engineer to help aid in the code review process.
- I will work to maintain smaller sets of changes in my PRs for the future. I understand the implications here of increasing code review labour. The nature of the initial state of this component meant that a large refactor was IMO, inevitable. 

**Major changes**
- Decomposes the main component into multiple smaller and focused files (including `.module.scss` files). Also renamed multiple components & functions to make more logical sense at a higher level.
- Allows for the consumer to pass a `links` object of their choosing for the generation of URLs and text. Updated test to cover this property being passed.

**Code Quality Improvements**
- Adds JSDoc comments
- Increases PropTypes usage
- Implements `lodash/get` for deeply nested properties
- Organizes imports
- Implements `src/helpers/isEnterKeyPress.js` as a separate helper function
- Implements `src/helpers/splitLastWord.js` as a separate helper function
- Increased spacing and color variable usage in `.scss` files
- Remove unnecessary conditions from `.map` usages
- Revise multiple css class definitions in JSX to use `array.join(' ')`
- State management improved to reduce re-rendering

**UI Improvements & Design QA**
- Clickable area for accordion toggle restricted to parent item
- Accordion links now properly tab indexed to allow for continuation of tabbing after expanding an accordion section, previously users were reset to the first mobile nav tab-able element.
[...]

PR #285, opened 2020-03-06
UniversalNavbarExpanded PropType comments & currentPage renames

c7hudson:
**Description:**
- Follow up to #279
- Renames some properties based on recent changes to helper name.
- Adds prop type comments to UniversalNavbarExpanded

**Screenshots:**
Screen Shot 2020-03-06 at 2 14 59 PM
Screen Shot 2020-03-06 at 2 15 44 PM
Screen Shot 2020-03-06 at 2 15 36 PM


c7hudson:
Not bumping version here as it's pretty low key stuff being changed.

PR #289, opened 2020-03-06
NoraCheckboxInput updates

stephaniel96:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1162798638929455/f)
- Adds capability to dynamically set `checked` on checkboxes. Blocked the MedExam Layout + Accordion Implementation effort, so this is related to that Asana task.

- https://github.com/getethos/ethos/pull/2514 pulls this in

### DIFF
--- a/src/components/CheckboxInput/CheckboxInput.js
+++ b/src/components/CheckboxInput/CheckboxInput.js
@@ -21,11 +21,13 @@ export const CheckboxInput = ({
   setFieldTouched,
   // TODO: pushed in from Form.js but not used here
   formTouched, // eslint-disable-line no-unused-vars
+  checked,
   ...rest
 }) => {
   const initialChecked = currentValue || initialValue || false
   const [touched, setTouched] = useState(initialValue ? true : false)
   const [isChecked, setIsChecked] = useState(initialChecked)
+
   const [getError, setError, , validate] = useErrorMessage(validator)
   const [doValidation] = useInputValidation({
     validate,
@@ -87,6 +89,11 @@ export const CheckboxInput = ({
       : `${styles.CheckboxInput}`
   }
 
+  /**
+   * If consumer's used the brute force `checked` prop, we prioritize that
+   */
+  const resolvedIsChecked = typeof checked !== 'undefined' ? checked : isChecked
+
   return (
     <>
       <label htmlFor={name} className={styles.root}>
@@ -98,7 +105,7 @@ export const CheckboxInput = ({
             data-tid={rest['data-tid']}
             onKeyPress={onKeyPress}
             disabled={disabled}
-            checked={isChecked}
+            checked={resolvedIsChecked}
             id={name}
             name={name}
           />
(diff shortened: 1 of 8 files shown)


src/nora/components/NoraCheckboxInput/NoraCheckboxInput.md
```diff
@@ -23,3 +23,49 @@ that accounts for the generally smaller font sizes we use in Nora.
   </a>
 </NoraCheckboxInput>
 ```
+
+Example with one checkbox setting the value of another
+
+```jsx
+import React, { useState } from 'react'
+
+function NoraCheckboxExample() {
```
stephaniel96: FYI @roblevintennis this was the example that I was unable to get working because there's no way to set the current checkbox value (and `currentValue` in `CheckboxInput` is the same as `initialValue`, it doesn't actually set the current value).

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -114,17 +121,35 @@ export const CheckboxInput = ({
...
+  /** low level prop used only by the form engine */
   formTouched: PropTypes.bool,
-  name: PropTypes.string.isRequired, // must be unique
+  /** Name of the field */
+  name: PropTypes.string.isRequired,
+  /** Required data-tid used as a unique id for targeting test selectors */
   'data-tid': PropTypes.string.isRequired,
+  /** Optionally sets a default value for the checkbox */
   initialValue: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+  /** Allows for brute force setting of whether checked or not. Usually,
```
roblevintennis: @stephaniel96 This is the compromise of this effort is to expose `checked` and have it take precedence over other means. Otherwise, we should let the checkboxes be dictated by the component itself.

I tried other means for a good couple of hours and this was the only approach I could get to work so 🤷‍♂

src/nora/components/NoraCheckboxInput/NoraCheckboxInput.md
```diff
@@ -23,3 +23,50 @@ that accounts for the generally smaller font sizes we use in Nora.
...
+  // Instead of a validator setting the state of the labs, we can also consider disabling the checkbox if one of the children are selected
+  return (
+    <>
+      <NoraCheckboxInput
+        name="labs"
+        data-tid="labs-checkbox"
+        checked={labSelected}
+        validator={(val) => {
+          setLabSelected(val)
+          if (!val) {
```
roblevintennis: Checking against `val` and not `labSelected` and `ceaSelected` was also key as it was causing a race condition.

src/nora/components/NoraCheckboxInput/NoraCheckboxInput.md
```diff
@@ -23,3 +23,50 @@ that accounts for the generally smaller font sizes we use in Nora.
...
+        }}
+      >
+        Labs
+        <NoraCheckboxInput
+          name="cea"
+          data-tid="cea-checkbox"
+          checked={ceaSelected}
+          validator={(val) => {
+            setCeaSelected(val)
+            if (val) {
```
roblevintennis: same here

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -87,6 +89,11 @@ export const CheckboxInput = ({
       : `${styles.CheckboxInput}`
   }
 
+  /**
+   * If consumer's used the brute force `checked` prop, we prioritize that
+   */
+  const resolvedIsChecked = typeof checked !== 'undefined' ? checked : isChecked
```
roblevintennis: So, my thought is that this is a purely additive change to the EDS checkbox component so there should be no risk of regression in terms of preserving previous functionality to existing usages across our Ethos apps. Put differently, if `checked` prop is not explicitly used by the consumer, everything else should be same as it was where `isChecked` dictates if the checkbox is checked or not.

src/components/CheckboxInput/CheckboxInput.js
```diff
@@ -114,17 +121,35 @@ export const CheckboxInput = ({
 }
 
 CheckboxInput.propTypes = {
+  /** low level prop used only by the form engine */
   formTouched: PropTypes.bool,
-  name: PropTypes.string.isRequired, // must be unique
+  /** Name of the field */
```
roblevintennis: We've learned that this sort of commenting helps with the output react-styleguidist generates in the EDS for props. We didn't do this in a lot of places, so I'm trying to just add them in as I encounter files organically.

src/components/CheckboxInput/CheckboxInput.js
roblevintennis: This comment is pretty important in relation to this PR specifically

src/nora/components/NoraCheckboxInput/NoraCheckboxInput.md
roblevintennis: Thanks for creating the test case Steph--I was able to work off it until I got the desired results for our use case. Working in isolation like this (outside of the main application) is probably 3x efficiency, and I'm glad we did it here :) We should continue to leverage this approach for EDS specific bugs / features / etc.

PR #296, opened 2020-03-10
FloatingDrawerContentSelector prop for Drawers

roblevintennis:
This adds the use outside click functionality including the floating action button (footer on the right side drawer) for https://github.com/getethos/ethos/pull/2528/files that was causing issues before the demo

### DIFF
--- a/src/hooks/a11y/useOutsideClickIgnoreSelector.js
+++ b/src/hooks/a11y/useOutsideClickIgnoreSelector.js
@@ -1,0 +1,38 @@
+import { useLayoutEffect } from 'react'
+
+/**
+ * Hook handles a click outside the targeted component
+ *
+ * @public
+ *
+ * @param {React.MutableRefObject<HTMLElement | null>} ref - the component's ref
+ * @param {() => void} handler - a callback that fires when an outside click is detected
+ *
+ * @return {void}
+ */
+function useOutsideClickIgnoreSelector(ref, ignoreSelector, handler) {
+  const handleClickOutside = (e) => {
+    console.log('floatingDrawerContentSelector. e.target: ', e.target)
+    let isIgnoredElement = false
+    if (e.target.classList && e.target.classList.contains(ignoreSelector)) {
+      isIgnoredElement = true
+    }
+    if (!isIgnoredElement && ref.current && !ref.current.contains(e.target)) {
+      handler.call(this, [e])
+    }
+  }
+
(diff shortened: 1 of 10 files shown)


src/hooks/a11y/useOutsideClickIgnoreSelector.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @public
+ *
+ * @param {React.MutableRefObject<HTMLElement | null>} ref - the component's ref
+ * @param {String} ignoreSelector - the selector to ignore when we consider outside clicks
+ * @param {() => void} handler - a callback that fires when an outside click is detected
+ *
+ * @return {void}
+ */
+function useOutsideClickIgnoreSelector(ref, ignoreSelector, handler) {
```
stephaniel96: @roblevintennis In Nora, the `ignoreSelector` is a style that we apply to the review button. I'm sure you've given this a lot of thought, but it feels slightly odd to me that we're going by a string, rather than a ref to a component or something. Is there any way that we could have an `ignoreRef` instead? Totally understand if that might not be possible or if it's overly complicated/unnecessary.

This is outside the immediate scope of this component, but I can see cases where we might want an array of `ignoreSelector` or `ignoreRefs`. In our case, Nora will need both a "Review Orders" button and "Submit Order" button, but since we're using the same styles for both, I think this will work for our case. However, it might be nice to add this functionality down the road!

roblevintennis: I started down the `ignoreRefs` road and the realized the renderer is dynamic for the floating action button so we essentially don't yet have access to the unrendered element yet at the time the Drawer.js code needs to initialize such a ref. So, I got creative and used a selector instead.

I think one stab I had set out to support an array of refs and as described above ran in to the not yet rendered issue; and so I think I was solving just to get it working and landed on this ignore selector solution. All to say, I think you're right that it would be even nicer if we supported and array of selectors 👍 

I'll try to work on the array of selectors approach when I get started today. Will also plan to add a few specs too. Thanks Steph, gonna set this to WIP until I get all that done.

stephaniel96: Gotcha! Thanks for the context. 👍

roblevintennis:
@stephaniel96 went ahead and re-requested a review since this is a fairly big change. It should support the array of selectors! 🙌

PR #298, opened 2020-03-13
Update IconLink to take in data-link-id

stephaniel96:
**Description:**

- While working on the left sidebar, we discovered that we need to be able to pass down a data- attribute to icon links to differentiate them since they are not unique based off of text.
- In other words, we want to be able to avoid this:

Image 2020-03-13 at 1 31 56 PM

- https://github.com/getethos/ethos/pull/2576

**Screenshots:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.131",
+  "version": "1.1.132",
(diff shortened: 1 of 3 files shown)


stephaniel96:
@roblevintennis confirmed that this works in the corresponding Nora PR. Accidentally bumped the version twice but otherwise this should be good to go!

PR #300, opened 2020-03-16
Removes the width style on select input

wafflehorse:
**Description:**
- Supports this ticket [Bug: Search bar component only shows last 9 characters of search typing](https://app.asana.com/0/1152139407849001/1162696936024777)
- There was a width put on the input style in order to override some of the default react-select styles. It's unclear to me why this had to be done and it's possible that I'm missing something. But when messing with the Select component variations with this change everything looks fine to me.
- Let me know if there are any ways I could test this further.

## Usages
- [x] **AuraDeSearchField - Physician Question**

### Before changes
Screen Shot 2020-03-17 at 9 20 12 AM

### After changes
Screen Shot 2020-03-17 at 9 20 56 AM

- [x] **Occupation Question**
### Before changes
Screen Shot 2020-03-17 at 9 15 52 AM
Screen Shot 2020-03-17 at 9 15 45 AM

### After changes
Screen Shot 2020-03-17 at 9 15 01 AM
Screen Shot 2020-03-17 at 9 15 20 AM


wafflehorse:
@roblevintennis I only see 2 usages in the monorepo and I've added screenshots above. Things look good to me!

PR #310, opened 2020-03-31
Tooltip Refactor for Customizability

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/0/1169062204585500/f)
- Refactor EDS Tooltip -- it's unclear how to opt out of the the information icons ('i' in circle) and we need to point to Nora's icon instead

https://github.com/getethos/ethos/pull/2752 will pull this in once ready

**Screenshots:**

Adds a Children section to the example. After reading through the code I figured out that you can in fact override the information circle icon by just passing in children and wrapping with the tooltip tag. But, there were no examples of how to do so. Also, I needed a way to override the `PopperContent` box styles and so I:

- renamed `.contentBox` to `.popperContentBox` and had to put it at top level of CSS Module so we can override it
- added a `popperBoxStyles` prop which does this overriding

image

I've already tested pulling into Nora and it's working:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.151",
+  "version": "1.1.152",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.152",
+  "version": "1.1.153",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.153",
+  "version": "1.1.154",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/src/components/Tooltip/Tooltip.md
+++ b/src/components/Tooltip/Tooltip.md
@@ -5,7 +5,6 @@
 To use a placeholder in its most basic form and with all defaults: provide two properties to the component, `label` and `details`
 
 ```jsx
-import React, { useRef, useState, useEffect } from 'react'
 import styles from './TooltipExamples.module.scss'
 ;<>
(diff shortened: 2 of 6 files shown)


src/components/Tooltip/Tooltip.module.scss
```diff
@@ -22,39 +22,6 @@
 }
 
 .popperContainer {
-  .contentBox {
```
roblevintennis: Having `. popperContainer .contentBox` is probably not needed given this is using CSS Modules, and, I only found the one usage within the tooltip component, so I elected to just move this to the top level and renamed it to `. popperContentBox`. That allows us to actually pass in an override prop `popperBoxStyles` without losing specificity battle.. Personally, I would use as flat as possible for all fo the CSS Modules stuff and the nested SCSS is overall a bit of an anti-pattern with this setup for just this reason (plus the fact it's generally best to keep specificity down to least possible).

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -110,6 +77,39 @@
   }
 }
 
+.popperContentBox {
```
roblevintennis: Same exact styles just moved to top level and renamed

src/components/Tooltip/TooltipExamples.module.scss
```diff
@@ -24,3 +24,9 @@
   align-items: center;
   height: 200px;
 }
+
+.CustomTipExample {
+  background-color: #0d1f40;
```
roblevintennis: I didn't bother using nora variables as this is just a custom example (even though this is the color we want 👅 )

src/components/Tooltip/Tooltip.md
```diff
@@ -5,21 +5,39 @@
...
+</>
+```
+
+
 ### `placement`
 
 Customize the **Starting** position of a Tooltip
 
 ```jsx
-import React, { useRef, useState, useEffect } from 'react'
```
roblevintennis: We don't need to import react and hooks the that component is already importing itself -- react-styleguidist will just pull in the component and it'd deps

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.151",
+  "version": "1.1.154",
```
roblevintennis: 🤷‍♂ it took me a few attempts forgetting this or that and that's why it's 3 version ^

PR #312, opened 2020-04-01
Modal typings

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/0/1169266212668985/f)
- EDS Modal doesn't have typescript typings in index.d.ts


src/components/index.d.ts
```diff
@@ -185,6 +185,34 @@ interface downstreamButtonProps {
...
+export declare const Modal: {
+  ({
+    ariaLabelledBy,
+    ariaDescribedBy,
+    children,
+    isOpen,
+    onDismiss,
+  }: {
+    /** Id of an element that labels the modal */
+    ariaLabelledBy?: string
```
stephaniel96, line 197: I was always confused about this spelling. Last I checked, `labeled` has only one `l`, but it seems like this is the standard haha

PR #313, opened 2020-04-03
Nora radios

roblevintennis:
**Description:**
[Asana](https://app.asana.com/0/1150068311310825/1167333682122591/f)
Nora modal

-

**Screenshots:**

image


src/nora/components/NoraRadioButtonGroup/NoraRadioButtonGroup.module.scss
```diff
@@ -0,0 +1,30 @@
...
+    height: var(--Nora-Font-15);
+  }
+
+  input:checked ~ aside {
+    border: 5px solid var(--NoraBlue-400);
+  }
+
+  /* Hacks: Coupled to base RadioButtonGroup's markup structure */
+  fieldset label {
+    /* 2nd label siblins onwards can't beat 3 classes with 3 elements so !important */
```
stephaniel96, line 14: This is somewhat confusing to me but it works! ¯\_(ツ)_/¯

roblevintennis: it's essentially same as:
`label + label` sibling selector, so if you had:
```html
<label>1</label>
<label>2</label>
<label>3</label>
```
only 2 and 3 will get the margin-top

roblevintennis: You can also use :not for this but the original I guess I used this technique and this is essentially overriding the 16px margin top with 12px so I had to use same approach

stephaniel96: Got it. Thanks for the explanation!

roblevintennis:
Ok, just pulled it into Nora and it appears to be working:

image


Merging

PR #315, opened 2020-04-08
Eds datagrid refactor

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/0/1169492014828504/f)
- Refactor EDS DataGrid to use Tom's Designs
- Remove the row sizes props and drive such customizations via the `--ariagrid-*` css custom properties defined by the consumer

**Screenshots:**

image


src/components/Grid/grid-example.module.scss
```diff
@@ -1,13 +1,61 @@
...
    * CSS Properites can take form of var(--possibly-set-var, --default-fallback-var)
    * allowing us to declaratively theme our grid!
    */
-  --ariagrid-grid-container-color: #272727;
-  --ariagrid-grid-border-bottom: #f5f5f5;
-  --ariagrid-header-border-bottom: #d8dcde;
   --ariagrid-row-hover-background: var(--BrandDuckegg);
-  --ariagrid-grid-header-font-color: var(--GraySecondary--opaque);
+
+  /* Main grid font color */
```
roblevintennis, line 11: My thought is we can literally copy these root properties over to Nora and then adds a media-query backed container to support 1440 / 1920 as the grid's container is width: 100% so it'll respect whatever parent element it's in.

PR #316, opened 2020-04-10
Data grid hotfix

roblevintennis:
This is just a follow up bugfix that I noticed on https://github.com/getethos/ethos/pull/2865 where I had duplicate props unintentionally

### DIFF
--- a/src/components/Grid/Column.module.scss
+++ b/src/components/Grid/Column.module.scss
@@ -26,7 +26,7 @@
   letter-spacing: var(--ariagrid-header-letter-spacing, 0.1em);
   /* since our header font size is likely different we need adjusted top/bottom padding */
   padding-top: var(--ariagrid-header-padding-top, var(--Space-18));
-  padding-bottom: var(--ariagrid-header-padding-top, var(--Space-18));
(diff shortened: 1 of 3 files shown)


src/components/Grid/Column.module.scss
```diff
@@ -26,7 +26,7 @@
   letter-spacing: var(--ariagrid-header-letter-spacing, 0.1em);
   /* since our header font size is likely different we need adjusted top/bottom padding */
   padding-top: var(--ariagrid-header-padding-top, var(--Space-18));
-  padding-bottom: var(--ariagrid-header-padding-top, var(--Space-18));
```
roblevintennis, line 29: bad copy pasta

src/components/Grid/grid-example.module.scss
```diff
@@ -42,7 +42,7 @@
   /* Header's padding top */
   --ariagrid-header-padding-top: 18px;
   /* Header's padding bottom */
-  --ariagrid-header-padding-top: 18px;
+  --ariagrid-header-padding-bottom: 18px;
```
roblevintennis, line 45: here too

PR #317, opened 2020-04-10
connects the nora textarea to the form engine

roblevintennis:
This is for getting the decline flow's components connected to eds form engine. This is the nora textarea one.

- https://github.com/getethos/ethos/pull/2879/files

**Screenshots:**
image

### DIFF
--- a/src/nora/components/NoraTextAreaInput/NoraTextAreaInput.js
+++ b/src/nora/components/NoraTextAreaInput/NoraTextAreaInput.js
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
-
+import useErrorMessage from '../../../hooks/useErrorMessage.js'
+import useInputValidation from '../../../hooks/useInputValidation.js'
 import { BaseTextAreaInput } from '../../../components/TextAreaInput/BaseTextAreaInput'
 import styles from './NoraTextAreaInput.module.scss'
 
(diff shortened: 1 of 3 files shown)


src/nora/components/NoraTextAreaInput/NoraTextAreaInput.js
```diff
@@ -17,7 +18,76 @@ export const NoraTextAreaInput = ({
...
+    if (onBlur) onBlur(ev)
+  }
+
+  /**
+   * If consumer's used the brute force `value` prop, we prioritize that
+   * (this is similar to what we're doing in CheckboxInput for `checked`)
+   */
+  const resolvedValue = typeof value !== 'undefined' ? value : val
+
+  // TODO: this indicates a field level error on the NoraTextAreaInput
```
roblevintennis, line 86: I figure I'll just leave this here and we can refactor once we know if/how we want to handle inline errors for Nora

src/nora/components/NoraTextAreaInput/NoraTextAreaInput.js
```diff
@@ -17,7 +18,76 @@ export const NoraTextAreaInput = ({
...
+  } = { ...rest }
+
   const classes = [styles.NoraTextInput, textClassName].join(' ')
+
+  // Set up validation hooks
+  const [getError, setError, , validate] = useErrorMessage(validator)
+
+  const [val, setVal] = useState(currentValue || initialValue || '')
+
+  const [touched, setTouched] = useState(initialValue ? true : false)
```
stephaniel96, line 37: What's the point of the logic here?

`NoraTextAreaInput` has a `value` prop which should be the default value. However, in this case, even when there is a default `value,` `touched` is `false` because `initialValue` is falsey.

Shouldn't we be checking `value` to double check whether this field is actually `touched`?

src/nora/components/NoraTextAreaInput/NoraTextAreaInput.js
```diff
@@ -17,7 +18,76 @@ export const NoraTextAreaInput = ({
...
+    // Also tell the form we've been touched
+    if (setFieldTouched) {
+      setFieldTouched(true)
+    }
+  }
+
+  // Initial Value aka prefilled—are considered "touched", but must prevalidate
+  // which will in turn update the internal form state as to their validity
+  useEffect(() => {
+    if (!!formChangeHandler && initialValue) {
```
stephaniel96, line 68: Similar to above, why is this logic using `initialValue` and not `value` or `resolvedValue`?

In the checkbox, we had to be backwards compatible. But shouldn't we be able to use only `resolvedValue` here?

roblevintennis: Updated

PR #323, opened 2020-04-15
change children prop type in Button from string to ReactNode

jarezvani:
**Description:**

- [Asana Task](https://app.asana.com/0/1152139407849001/1167233174456285)
- 

-

**Screenshots:**
- added section to Styleguide

Screen Shot 2020-04-15 at 3 46 09 PM

### DIFF
--- a/src/components/Button/Button.md
+++ b/src/components/Button/Button.md
@@ -109,6 +109,36 @@ Set of Design-approved public buttons.
 <Button.Unstyled backArrowIcon>Button.Unstyled + backArrowIcon</Button.Unstyled>
 ```
 
+#### React Nodes as Button children
+
+```jsx
+  <div style={{ backgroundColor: 'var(--BrandForest)', padding: 20 }}>
+    <strong style={{ color: 'white' }}>
+      Children can be ReactNode, Array&lt;ReactNode&gt;, string, number, undefined, null, boolean
+    </strong>
+    <br />
+    <br />
+    <Button.Medium.WhiteOutline>
+      {/*  */}
+      {[
+        <div
+          key="top-div"
+          style={{backgroundColor: 'rgba(0, 0, 0, 0.1)', height: '30%', padding: '0.5rem'}}
+        >
+          Top Div
+        </div>,
+        <div
+          key="bottom-div"
+          style={{backgroundColor: 'rgba(0, 0, 0, 0.2)', height: '30%', padding: '0.5rem'}}
+        >
+          Bottom Div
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.md
```diff
@@ -109,6 +109,36 @@ Set of Design-approved public buttons.
...
+        >
+          Top Div
+        </div>,
+        <div
+          key="bottom-div"
+          style={{backgroundColor: 'rgba(0, 0, 0, 0.2)', height: '30%', padding: '0.5rem'}}
+        >
+          Bottom Div
+        </div>,
+        <br key='br'/>,
```
roblevintennis, line 135: Why do you need this key?

src/components/Button/Button.md
```diff
@@ -109,6 +109,36 @@ Set of Design-approved public buttons.
 <Button.Unstyled backArrowIcon>Button.Unstyled + backArrowIcon</Button.Unstyled>
 ```
 
+#### React Nodes as Button children
+
+```jsx
+  <div style={{ backgroundColor: 'var(--BrandForest)', padding: 20 }}>
```
roblevintennis, line 115: I see what you're doing here and I like the verification. I don't know if you need it in the documentation though. I bet you could convert this to a jest snapshot quite easily and still maintain the verification

src/components/Button/Button.md
jarezvani: I'm passing it in a `Array<ReactNode>` so it says it needs a key since it's in a list

src/components/Button/Button.test.js
```diff
@@ -56,5 +56,23 @@ describe('Button', () => {
...
+              null,
+              undefined,
+              3,
+              'string',
+              <React.Fragment key="fragment">Fragment with key</React.Fragment>,
+            ]}
+          </Button.Medium.Black>
+        )
+        .toJSON()
+      expect(tree).toMatchSnapshot()
```
roblevintennis, line 75: oh yeah, this is exactly what I was talking about 👍 -- I'd say the markdown is a great local test to do and I'm glad you did, but maybe we should prune it from the PR

src/components/Button/Button.md
jarezvani: cool yeah, I was on the fence about it, but figured I'd push it as a kind of POC; I can remove it

src/components/Button/Button.test.js
jarezvani: will do

PR #324, opened 2020-04-16
Kabob Menu stuff

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1170387956248380/f)
- Kabob Menu 

- [ ] Have you exported it from the [index.d.ts](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts) so Typescript users can consume it? Added a test 

### Not yet. I'd prefer to do in another follow up PR for the typings and then also pull into Nora and determining if said types and usage are meeting our needs.

Not yet, again I'd like to do this in a follow up PR

**Screenshots:**

Hovered
image

Hovering over the currently "selected" item:

image


src/components/Popover/Example.module.scss
```diff
@@ -0,0 +1,5 @@
+.Container {
+  --async-active-color: #b8c7d6;
```
roblevintennis, line 2: These are arbitrary and just meant to show you can customize via css properties

src/nora/components/KabobMenu/KabobMenuContainer.md
```diff
@@ -0,0 +1,39 @@
+```jsx
+import styles from './KabobExample.module.scss'
+import { Body2, Spacer } from '../../../components/index'
+
+const ExampleKabobMenuConsumer = () => {
+  const [isOpen, setIsOpen] = React.useState(false)
+  const [activeOption, setActiveOption] = React.useState(0)
```
sriram-ethos, line 7: nit: no need for the 'activeOption' in this example? or did you mean to use it in the example?

src/components/Popover/Options.md
```diff
@@ -0,0 +1,90 @@
...
+  const optionsRefs = Array(items.length)
+    .fill(0)
+    .map(() => React.createRef())
+
+  const handleOnChange = (item) => {
+    setLastSelected(item.name)
+  }
+
+  const setSelectedAndActiveOptions = (index) => {
+    setSelectedOption(index)
```
sriram-ethos, line 43: The difference between 'active' and 'selected' is not very clear when reading this and the other files.  Maybe it can be made clearer - either by renaming one of them to be more specific (active == ?) or updating the readme with some explanation?

src/nora/components/KabobMenu/KabobMenu.js
```diff
@@ -0,0 +1,46 @@
...
+  const containerClasses = className
+    ? `${styles.Container} ${className}`
+    : styles.Container
+
+  return (
+    <div className={containerClasses} onKeyDown={onKeyDown}>
+      <button className={styles.Box} onClick={onClick}>
+        <KabobIcon />
+      </button>
+      {children}
```
sriram-ethos, line 32: Do you envision the 'KabobMenu' taking other forms as well apart from the 'children' being just a 'list' of items? Since it is a 'menu', wouldn't it always have a 'list of items' that need to be displayed as a 'list'.

src/components/Popover/Options.md
sriram-ethos: Is 'active' == 'highlighted'?

src/nora/components/KabobMenu/KabobMenu.js
roblevintennis: Perhaps it will in practice be a list, but, having it take in `{children}` like this allows us to use the popover set of components in a flexible way e.g. we could customize it with another list component if we elect to; it also decouples the two as opposed to the Kabob Menu needing to know about it's children or how to deal with dropdown options as such. In fact, the popover stuff was teased out from the AsyncTypeahead component which was in fact having this very problem until I refactored. Now I get to use those dropdown options aka Popover component in as many places I wish and get a11y, styling, etc., for free.

src/components/Popover/Options.md
roblevintennis: I think you're questions and the confusion totally make sense. Yes to the second question, it's the "highlighted" visually, but really it's the focused item since these are used to support the a11y keyboard navigability.

It's my Sunday and I have my kids, but I did just add a commit which I think best addresses this as I hadn't documented these props!

```js
Option.propTypes = {
  /** The currently active option item. This is the item that is currently
   * being focused on (but not necessarily selected yet!) */
  currentActive: PropTypes.number,
  /** The option item's index in the options */
  itemIndex: PropTypes.number,
  /** The data object with the data for this option via the `dataKey` */
  item: PropTypes.object,
  /** The key to access via `item` e.g. `item.name` */
  dataKey: PropTypes.string,
  /** The currently selected option item. Unlike an active item, this signifies
      that the user has actually selected or chosen this item. Of course an
      item can be both active and selected. */
  selectedOption: PropTypes.number,
  /** Callback that gets called once an option is selected */
  setSelectedAndActiveOptions: PropTypes.func,
  /** Callback that gets called on change */
  onChange: PropTypes.func,
}
```

I still like active better then highlighted, so I'd like to keep the variable names as is with hopes that improved docs takes care of this being unclear. Thanks for the feedback Sriram!

src/nora/components/KabobMenu/KabobMenuContainer.md
roblevintennis: Yup, thanks Sriram, fixed in:
67433acdc0448c827f8c68749b8c1045f3d106d5

src/nora/components/KabobMenu/KabobMenu.js
sriram-ethos: Makes sense.

src/components/Popover/Options.md
sriram-ethos: I also think 'active' is better and I think your comment helps in understanding this.

roblevintennis:
Thanks for the review @sriram-ethos -- it was helpful to get your perspective on this 💯 🙏 gonna merge it now

PR #326, opened 2020-04-21
Nora kabob types version bump etc

roblevintennis:
This is a part of https://app.asana.com/0/1150068311310825/1170387956248380/f

I just used the dashboard's list view to test this out (but it won't actually go there when we use it -- it's for some of our detail view's decision tables). But:

image

I still have some work to do on making this fully work with the data grid, but I have that follow up tracked on the Asana. I'd like to iterate so I'm looking to merge what I have here if possible.

### DIFF
--- a/src/components/Popover/Options.js
+++ b/src/components/Popover/Options.js
@@ -53,11 +53,11 @@ export const Options = ({
 }
 Options.displayName = 'Options'
 Options.propTypes = {
-  activeOption: PropTypes.number,
-  dataKey: PropTypes.string,
-  entities: PropTypes.array,
-  onChange: PropTypes.func,
-  optionsRefs: PropTypes.array,
-  selectedOption: PropTypes.number,
-  setSelectedAndActiveOptions: PropTypes.func,
+  activeOption: PropTypes.number.isRequired,
+  dataKey: PropTypes.string.isRequired,
+  entities: PropTypes.array.isRequired,
+  onChange: PropTypes.func.isRequired,
+  optionsRefs: PropTypes.array.isRequired,
+  selectedOption: PropTypes.number.isRequired,
+  setSelectedAndActiveOptions: PropTypes.func.isRequired,
 }
--- a/src/nora/components/KabobMenu/KabobMenu.js
+++ b/src/nora/components/KabobMenu/KabobMenu.js
@@ -45,7 +45,9 @@ export const KabobMenu = ({
 }
 
 KabobMenu.propTypes = {
-  /** `tabIndex` - Used if data grid wants to put focus */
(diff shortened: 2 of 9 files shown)


src/components/Popover/Options.js
```diff
@@ -53,11 +53,11 @@ export const Options = ({
...
-  optionsRefs: PropTypes.array,
-  selectedOption: PropTypes.number,
-  setSelectedAndActiveOptions: PropTypes.func,
+  activeOption: PropTypes.number.isRequired,
+  dataKey: PropTypes.string.isRequired,
+  entities: PropTypes.array.isRequired,
+  onChange: PropTypes.func.isRequired,
+  optionsRefs: PropTypes.array.isRequired,
+  selectedOption: PropTypes.number.isRequired,
+  setSelectedAndActiveOptions: PropTypes.func.isRequired,
```
roblevintennis, line 62: My usage(s) use all these props and I don't see it really being functional without them, so I've err'd on side of "is required". I suppose we can back that off in the future if we find use cases like we did for `labelCopy` in other components.

src/nora/components/KabobMenu/KabobMenu.js
```diff
@@ -19,24 +19,36 @@ const KabobIcon = () => {
...
+  const tabIndexResolved = tabIndex ? tabIndex : null
+  const focusRefResolved = focusRef ? focusRef : null
   return (
-    <div className={containerClasses} onKeyDown={onKeyDown}>
-      <button className={styles.Box} onClick={onClick}>
+    <>
+      <button
+        ref={focusRefResolved}
+        tabIndex={tabIndexResolved}
+        className={styles.Box}
```
stephaniel96, line 36: Was your thinking that we'd have an enforced style for the Kabob Menu for now, and if down the road we need some customization of styling (for Nora or other products) we can refactor or potentially create a new component?

roblevintennis: No, there's been quite a bit of difficulty in just getting this to work and so I think this was an oversight on my part! I think the best and easiest approach at this point is for me to replace some of the hard coded values with the conditional CSS props pattern that's been getting introduced elsewhere like the data grid and what not e.g. `var(--override-value, var(--fallback-value))`. Thanks for noticing this Steph!

I'm going to just piggyback that in here before merging.

stephaniel96: Awesome! Thanks for the explanation :)

roblevintennis: Thank you Steph, nice catch!

I've just pushed up with defs like `--kabob-*` which will achieve ^

PR #327, opened 2020-04-21
Data grid's keydown listener competing with popover listener bugz

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1170387956248380/f)
- Fix issue with kabob and data grid competing for focus

image

- Fix issue where when you navigate down from Grid after already having focus on the Kabob's trigger button, it would open that popover (even though we're now on the next trigger button down by 1 row)

image

___

**Manual Testing:**
I have a dirty Nora branch where I've been verifying all these changes in the data grid. This fixed the two bugs 🙌

### DIFF
--- a/src/components/AsyncTypeahead/AsyncTypeahead.js
+++ b/src/components/AsyncTypeahead/AsyncTypeahead.js
@@ -137,6 +137,7 @@ export const AsyncTypeahead = ({
         onClick: setShow,
         onKeyDown: (ev) =>
           handlePopoverNavigation({
+            autoopenOnFocus: true,
             ev,
             items: entities,
             onChange,
--- a/src/nora/components/KabobMenu/KabobMenuContainer.js
+++ b/src/nora/components/KabobMenu/KabobMenuContainer.js
@@ -73,11 +73,13 @@ export const KabobMenuContainer = ({
     const elements = getFocusableElements(popoverTrapRef.current)
     if (isOpen && elements.length > 0) {
       elements[0].focus()
-      document.body.addEventListener('keydown', handleOnKeyDown)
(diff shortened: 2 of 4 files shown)


src/components/AsyncTypeahead/AsyncTypeahead.js
```diff
@@ -137,6 +137,7 @@ export const AsyncTypeahead = ({
         onClick: setShow,
         onKeyDown: (ev) =>
           handlePopoverNavigation({
+            autoopenOnFocus: true,
```
roblevintennis, line 140: Context:
As I'd originally only used the popovers on the async typeahead component, I had this sort of "feature" where if you focused on the search typeahead component it would auto open (as expected for a search typeahead component, but not for the kabob button of course!) -- so this just rewired the usage to explicitly opt in (it defaults to falsy otherwise)

src/nora/components/KabobMenu/KabobMenuContainer.js
```diff
@@ -36,27 +42,55 @@ export const KabobMenuContainer = ({
...
+      activeOption: activeOption,
+      setActiveOption: setActiveOption,
+      scrollItemIntoView: scrollItemIntoView,
+      showPopover: isOpen,
+      setShowPopover: setIsOpen,
+      setSelectedAndActiveOptions: setSelectedAndActiveOptions,
+    })
+  }
+
+  const popoverTrapRef = React.useRef()
```
roblevintennis, line 70: This whole section basically:

- puts focus on the first focusable element in the popover
- adds listener if we've opened the popover
- removes the listener if we tear down the component (if the ref isn't removed already which is usually the case)

PR #329, opened 2020-04-23
CMS Estimate experiment in UniversalNavbarExpanded

c7hudson:
**Description:**
- Would like to merge by EOD April 23rd if possible.
- Diff is largely made up of snapshot update! Changes are smaller than they look.
- [Asana Task](https://app.asana.com/0/1116481661798430/1166819983224779/f)
- Need some Optimizely Web selectors and additional copy in the navbar, for running some CMS side experiment targeting on all pages.

**Is this a new component? Please review these reminders:**
Not a new component.
- [x] Followed the checklist at the bottom of the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute) guide?

- https://github.com/getethos/cms/pull/2930 (original estimate widget PR - being broken into multiple smaller PRs)

**Screenshots:**
Screen Shot 2020-04-23 at 1 51 58 AM

### DIFF
--- a/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
+++ b/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
@@ -6,7 +6,7 @@ import LogoNotAnimated from '../UniversalNavbar/assets/ethos-logo-black.js'
 import { AccountIcon, SearchIcon } from '../UniversalNavbar/assets/icons.js'
 
 // EDS core components
-import { Layout } from '../index'
+import { Layout, Caption2 } from '../index'
 
 // UniversalNavbarExpanded simple siblings
 import CtaButton from './CtaButton'
@@ -34,6 +34,7 @@ import styles from './UniversalNavbarExpanded.module.scss'
  * @param {object} LinkComponent - Agnotistic Reach and React Router Link (ex. Gatsby's <Link>)
  * @param {string} logoHref - Href for the logo
  * @param {object} links - URLs and text
+ * @param {boolean} estimateExperiment - enable the estimate experiment button/copy
  *
  * @return {JSX.Element}
  */
@@ -44,6 +45,7 @@ const UniversalNavbarExpanded = ({
   logoHref,
   trackCtaClick,
   links,
+  estimateExperiment,
 }) => {
   const BELOW_ACCORDION_LINKS = [links.CTA, links.ACCOUNT, links.SEARCH]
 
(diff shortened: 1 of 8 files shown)


src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
```diff
@@ -95,3 +95,14 @@ $laptop-range-small-end: 929px;
     display: none;
   }
 }
+
+.estimateCopyOptimizely {
```
stephaniel96, line 99: Unrelated, but have we decided to use camel case instead of pascal case? I've still been trying to use pascal case but more than happy to switch to camel case if we've decided we want that going forward 👍

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
```diff
@@ -166,6 +180,8 @@ UniversalNavbarExpanded.propTypes = {
       })
     ),
   }).isRequired,
+  /** Estimate copy experiment on optimizely */
+  estimateExperiment: PropTypes.bool,
```
stephaniel96, line 184: Would it be possible to have this component take an optional `children` or `experimentChildren` and leave the component content, design, and optimizely tracking up to the consumer of `UniversalNavbarExpanded`?

Reasoning behind this is that it seems like we're slowly leaning towards allowing consumers of EDS to specify className and other properties that were previously hardcoded. The UniversalNavbar is still pretty opinionated, but it could be nice to slowly refactor it to support more behaviors.

c7hudson: @stephaniel96 

If I had more time I would be willing to make the switch to passing in children. We're trying to get this experiment turned on tomorrow and I've still got some loose ends CMS side to tidy. 

This `Expanded` version is already specific to CMS only and this `estimateExperiment` prop is still optional, should another consumer need to use the navbar and not worry about specifying a value here.

I'd be willing to make a growth backlog ticket here, to address your request when I have more bandwidth, so that future consumers can pass in children or classes. 

This experiment code should only be temporary and removed once we see a winner in Optimizely.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
c7hudson: I'm not sure if we've documented a standard for EDS regarding casing. We should if we haven't. Good catch.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
stephaniel96: Thanks for your explanation! I didn't realize that this was a specific `Expanded` version, so that's an extra compelling reason 👍

PR #331, opened 2020-04-28
Nora Button Inputs

roblevintennis:
…e for NoraButtonInput

**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1171508632702019/f)
- Nora Button Inputs

**TODOs:**
- [x] Add hover states with default but also a prop that can be passed in.
- [x] Refactor the `onClick` so it calls back to consumer
- [x] Fix trash can is misaligned issue
- [x] re-xport and bump eds
___

**Screenshots:**

image

In Nora:

image

https://github.com/getethos/ethos/pull/3142


roblevintennis:
@sriram-ethos I haven't attempted to pull this in to Nora but in case you needed this for MIB Codes it's now available. If you can wait, I can set it up from Nora side tomorrow.

src/nora/components/NoraButtonInput/NoraButtonInput.js
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <div className={styles.NoraButtonInputContainer}>
+      {side === 'right' && (
+        <NoraTextInput className={styles.InputLeft} {...cloned} />
+      )}
+      <button
+        className={getButtonStyles()}
+        tabIndex="-1"
+        onClick={() => {
+          console.log('got clicked...')
```
roblevintennis, line 49: Just noticed this. I'll have to work some more on this tomorrow. I think we'll need to have this as a prop and call it back if not `buttonDisabled` prop

sriram-ethos:
This looks good to me so far.  One thing I'd probably add to this is probably the `hover` behavior - the `TextInput` has a hover behaviour and I'd think the `button` should also have a hover behavior, where the icon changes from gray to blue or something like that.  Does this make sense?

src/components/index.d.ts
```diff
@@ -1308,6 +1308,46 @@ export declare const NoraCheckboxInput: {
     props: any
   }
 }
+export declare const NoraButtonInput: {
+  ({
+    iconPrefix,
+    iconName,
+    side,
+    buttonDisabled,
+    // NoraTextInput props are passed through
```
sriram-ethos, line 1317: The 'pass through' parameters for `TextInput` - is there a need to enumerate them one by one or can we do a spread of the props to the `TextInput`, similar to how you've done in other derived input components.

roblevintennis: In this d.ts file, my thought was that explicitly defining these would provide the type hinting in Typescript and our IDE's.

sriram-ethos: The more type hinting, the better :)

src/nora/components/NoraButtonInput/NoraButtonInput.module.scss
```diff
@@ -0,0 +1,41 @@
...
+.ButtonLeft {
+  composes: Button;
+  margin-right: -1px;
+}
+.ButtonRight {
+  composes: Button;
+  margin-left: -1px;
+}
+
+.Disabled {
```
roblevintennis, line 39: Per @sriram-ethos other comment regarding hover, I agree that it's missing here. I think we should do some sort of `:not(.Disabled):hover` or even negate the hover rules in .Disabled itself (might be more straight forward).

roblevintennis: Actually, I think since we do `pointer-events: none` and `.Disabled` is removing button interactively, all above is probably not even needed. Put differently, my guess is we just add the hover functionality, and when the button's disabled, it will "just work" and no hover rules will be able to get applied.

roblevintennis:
I've created a TODOs section of the description. I'll work off that later when I start my workday :)

src/nora/components/NoraButtonInput/NoraButtonInput.module.scss
sriram-ethos: I agree.  I think its better to start from scratch (similar to other props) as the behavior is not the same as the TextInput.

PR #333, opened 2020-05-05
Add aria-label and data-tid props

anbright:
**Description:**

- Based on a PR comment (https://github.com/getethos/ethos/pull/3194#pullrequestreview-405364329), adds an `aria-label` to the `Select` component for a11y
- Adds a `data-tid` prop to `Select`
- Adds snapshot tests

Reference but does not depend on these changes
- https://github.com/getethos/ethos/pull/3194

**Screenshots:**
image
image

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -342,6 +342,7 @@ export declare const Select: {
     title?: string
     className?: string
     isCreatable?: boolean
+    'data-tid'?: string
   }
   defaultProps: {
     classNamePrefix?: string
--- a/src/components/index.tests.tsx
+++ b/src/components/index.tests.tsx
@@ -383,8 +383,14 @@ class SelectTest extends React.Component<any, any> {
           options={options}
           isAsync={true}
           isCreatable={true}
+          data-tid="foo"
         />
-        <Select onChange={onSelected} options={options} isAsync={true} />
+        <Select
+          onChange={onSelected}
+          options={options}
+          isAsync={true}
+          data-tid="foo"
+        />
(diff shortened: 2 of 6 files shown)


anbright:
> Whoa, thanks! 🙏 If you don't mind, can you update the index.d.ts typings and check if index.tests.tsx could use it, and then run `yarn test:types` to verify it please? Otherwise I believe this will throw a typescript error for consumers like Nora.

Updated in 6b8f3d0 and updated test in 3b7de57

src/components/index.d.ts
```diff
@@ -342,6 +342,7 @@ export declare const Select: {
     title?: string
     className?: string
     isCreatable?: boolean
+    'data-tid'?: string
```
roblevintennis, line 345: thanks!

src/components/index.tests.tsx
```diff
@@ -383,8 +383,14 @@ class SelectTest extends React.Component<any, any> {
           options={options}
           isAsync={true}
           isCreatable={true}
+          data-tid="foo"
+        />
+        <Select
```
roblevintennis, line 388: I should mention to you (and I've said this in various forums so you may or may not have heard it) but Select has an issue in that it doesn't work with the form engine validation properly. In practice, this hasn't been an issue for the most part because we don't usually need to validate what's selected per se, but, if you do, please be aware this is a "known bug" if not the most serious one in EDS cc / @anbright

anbright: Alright, thanks for the heads up!

PR #334, opened 2020-05-06
Stepper

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1147552137426708/1171609519486697/f)
- Creating a navigation component -- it's just a dumb vertical stepper; state is pushed in from outside.

_Note that at this point it isn't interactive per the state of the funnel at the moment. When we build in user interaction we'll likely use `<button>` elements and change some of the a11y -- my research indicated we need to not use `<nav>` until it's interactive fwiw._

This is going to be used for something like the following:

image

- Corresponding PR: https://github.com/getethos/ethos/pull/3231
Seems to be working from the app shell:
image

**Screenshots:**

These are tests within the EDS:

image

image

image

I also seem to be able to arbitrarily add steps and it grows dynamically as expected:

image


src/components/index.tests.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Accordion } from './index'
+import { Accordion, Stepper } from './index'
```
stephaniel96, line 2: Is there a reason you've opted to import `Stepper` and `Accordion` in the same line?

roblevintennis: Lol, no 🤦 I'll fix that thanks Steph

roblevintennis: I'm actually suspicious that vscode auto did that lol

roblevintennis: fixed in v1.1.187

PR #335, opened 2020-05-06
Stepper p2 and type2 document public props usage

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1147552137426708/1171609519486689/f)
- Updates the Stepper to use the Type 2 system
- Adds some documentation to EDS on Type 2 system examples making it super clear that you can use any of the type base "public props"

- https://github.com/getethos/ethos/pull/3237

**Screenshots:**
image

image

Did same as above for all the type 2 Title* component examples too

### DIFF
--- a/src/components/Stepper/Stepper.js
+++ b/src/components/Stepper/Stepper.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Icon } from '../index'
+import { COLORS, Caption2, Icon } from '../index'
 import PropTypes from 'prop-types'
 import uuidv4 from 'uuid/v4'
 import styles from './Stepper.module.scss'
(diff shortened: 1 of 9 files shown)


src/components/TypeBase/TypeBase.js
```diff
@@ -168,6 +168,7 @@ export const TypeFoundry = (privateProps) => {
 
   // TODO: figure out if downstream or upstream is the correct nomenclature here
   const PublicTypeComponent = (downstreamProps) => {
+    console.log('downstreamProps props in type foundry: ', downstreamProps)
```
stephaniel96, line 171: Perhaps we can remove this?

roblevintennis: yup...will be a question New version: 1.1.189 coming soon to reflect

src/components/Stepper/Stepper.js
```diff
@@ -18,13 +18,19 @@ export const Stepper = ({ steps }) => {
...
+    if (step.status == 'complete' || step.status == 'active') {
+      return (
+        <Caption2.Regular400 color={COLORS.GRAY_PRIMARY}>
+          {step.title}
+        </Caption2.Regular400>
+      )
     }
-    return styles.Title
+    return (
+      <Caption2.Regular400 color={COLORS.GRAY_STROKE_AND_DISABLED}>
```
c7hudson, line 30: minor non blocking nit - could set color in a variable and reduce this to just `<Caption2.Regular400 color={colorVar}>` if everything else is the same.

PR #337, opened 2020-05-11
Refactor selects

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1136962714401929/1155914817079659/f)
- EDS Select Initial Values and Validation
- Monorepo PR: https://github.com/getethos/ethos/pull/3312

- [x] is there a form engine example within the Select field example?
- [x] Is single select working?
- [x] Is multi select also working?

- [x] Figure out which of the shared fields is wrapping this and verify in funnel

- frontend/src/main/InsuranceApp/term/Apply/Greenhouse/AuraDe/AuraDeQuestionUi.js for AuraDeSearchQuestion case (Eric and Jas):

```

          <AuraDeSearchField
            {...fieldProps}
            cacheOptions
            isAsync
            isMulti
            loadOptions={loadOptions}
          />
```
For above, it looks like the "**What is your occupation?**" aura question uses this and is an interesting usage in terms of it's async. The Doctor question also appears to use Select indirectly but via _AuraDeSearchComponent_

- underwriting/nora/frontend/src/features/dashboard/Toolbar.tsx for the dashboard filters in toolbar (Steph)

- frontend/src/main/InsuranceApp/term/Apply/Greenhouse/Checkout/components/PolicyDetailsCard.js appears to use following which wrap <Select /) (Drew):

    - frontend/shared/components/Dropdown/CoverageDropdown.js 
    - frontend/shared/components/Dropdown/TermDropdown.js

- frontend/src/main/InsuranceApp/PreInterview/Questions/components.js uses in pre interview aka Starboy getSelectComponent (James) for states dropdown, birthCountry dropdown etc. -- all appear to be single selects

- [x] Dev QA (see screengrabs at bottom of this PR)

- [x] Does `defaultValue` still work properly?

**Screenshots:**

Single select tests:

initial state form is invalid so submit button is reflecting this will disabled button:
image

selecting an invalid option:

image

selecting an valid option:

image

___ 

Here's multi which I've added as a second field to the form engine example:

Valid selection (in this case I have a validator setup to ensure 2+ items selected):

image

Invalid (only 1 selected):
image

Edge case -- user selects some multi items, then clears them out -- still results in validation working properly:

image

Improves the docs for the single vs. multi validation complexity:

image

Also adds a form engine example with even more explanation!

image

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 // See React-Select -- https://github.com/JedWatson/react-select for documentation
 // on usage, Async configuration, etc.
@@ -6,6 +6,8 @@ import ReactSelect from 'react-select'
 import ReactSelectAsync from 'react-select/async'
 import ReactSelectAsyncCreatable from 'react-select/async-creatable'
 import ReactSelectCreatable from 'react-select/creatable'
+import useErrorMessage from '../../hooks/useErrorMessage.js'
+import { InputLabel } from '../InputLabel'
 
 import styles from './Select.module.scss'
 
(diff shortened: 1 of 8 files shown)


roblevintennis:
@wafflehorse this is still in wip draft because I need to update the typescript types and what not, but can you have a preemptive look at this? What seemed to be the most confusing part of this was actually the react-select libraries [state manager](https://react-select.com/props#statemanager-props) which calls our on change handler with shape:
```
One of<
Object,
Array<Object>,
null,
undefined
>
```

Because of all that, I feel like this is a sensible comprimise:

1. continue to use a string value for the single select validator callback case
2. for multi, normalize their array to just an array of values which feels the closest to what we normally do in EDS validation, to what's being done for the single select (otherwise, if we just pass back the react-select data structure it's like `[{label: 'abc', value: 'la la'}, {label: 'xyz', value: 'yoyo'}]` and that is such a departure from our current validators I've elected to prefer and array of string values; the array is still a change from before, but at least it's not a huge one-off one.

src/components/Select/Select.js
```diff
@@ -1,29 +1,106 @@
...
+  formTouched,
+  labelCopy,
+  name,
   ...rest
 }) => {
+  const resolvedValidator = validator ? validator : () => ''
+  const [getError, setError, , validate] = useErrorMessage(resolvedValidator)
   const compactClass = isCompact ? styles.Compact : ''
+
+  // According to the react-select https://react-select.com/props#statemanager-props docs
```
wafflehorse, line 41: Sweet sweet documentation!

roblevintennis:
Thanks @wafflehorse 

> multi select and single select being different components

Yeah, I was thinking of that and honestly trying to find the easiest least changes approach since this is already in use. I suppose I'm fearing the added work in backfilling haha. If you don't mind, I'll see how far I can get leaving 'em together for now and keep the idea in my "back pocket" as they say :)

roblevintennis:
# Dev QA

image


image

jameshawkinsjr:
@tkarssli @anbright this will allow us to run validations on the address page, so when someone selects `NY` we can say we don't offer policies in that state.

roblevintennis:
@jameshawkinsjr 

>@tkarssli @anbright this will allow us to run validations on Select components (i.e. address page or if we wanted to convert the citizen question back to a dropdown), so when someone selects NY we can say we don't offer policies in that state.

Another typical UX that comes to my mind is the case where you require the user to have explicitly made a choice from a dropdown (sometimes legal even requires this), you can set the first option to `- Select -` or something to that affect, and then fail if they haven't actually chosen a real option value. Might be worth spreading the word to PM's too since they probably just work around this limitation today. I might broadcast that as well :)

PR #341, opened 2020-05-18
Ab/add aria label

anbright:
**Description:**

- [Asana Task](https://app.asana.com/0/1116481661798430/1172907631940805)
- Adds an `aria-label` to the `TextInput` which is a `critical` failing check on the Microsoft Accessibility audit

- https://github.com/getethos/ethos/pull/3393

### DIFF
--- a/src/components/TextInput/TextInput.js
+++ b/src/components/TextInput/TextInput.js
@@ -111,6 +111,7 @@ function PrivateTextInput({
         onBlur={onBlur}
         value={value}
         data-tid={rest['data-tid']}
+        aria-label={name} // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute
       />
       {getError(currentError, touched)}
     </>
--- a/src/components/EmailInput/__snapshots__/EmailInput.test.js.snap
+++ b/src/components/EmailInput/__snapshots__/EmailInput.test.js.snap
@@ -18,6 +18,7 @@ Array [
     }
   />,
   <input
+    aria-label="the-email-input-example"
     className="TextInput"
     data-tid="the-email-input"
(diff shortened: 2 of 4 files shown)


src/components/UniversalNavbar/TransformingBurgerButton/TransformingBurgerButton.js
```diff
@@ -16,6 +16,7 @@ const TransformingBurgerButton = ({
         className={styles.iconWrapper}
         role="button"
         tabIndex="0"
+        aria-label="Menu"
```
roblevintennis, line 19: Please check if this affects the types in index.d.ts

src/components/TextInput/TextInput.js
```diff
@@ -111,6 +111,7 @@ function PrivateTextInput({
         onBlur={onBlur}
         value={value}
         data-tid={rest['data-tid']}
+        aria-label={name} // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute
```
roblevintennis, line 114: Is this strictly an internal call to input? Perhaps the index.d.ts does NOT need updating? 🤔

anbright:
Yes, they are getting passed through `...rest`. I added an `aria-label` to the NoraTextInput test and ran yarn test:types to confirm it was working.

src/components/TextInput/TextInput.js
anbright: Yes, it should be an internal call

> I think my initial take was that we'd need to update the index.d.ts but it looks like these aria-label attributes may be just getting passed through in a `...rest` or similar prop. Please confirm and verify it doesn't break typescript consumers before merging otherwise LGTM -- thanks!

PR #342, opened 2020-05-19
Date inputs

roblevintennis:
**Description:**

- [Asana Task](https://app.asana.com/0/1147552137426708/1175489927710786/f)
- Date input (masked) UI Components

### Context

We probably should not have called the birthdate input `BirthdateInput` as it's mostly just a generic date input. It so happens we actually need a more generic date input for things like "Last Seen" question which is in `MM/YYYY` format and, of course, has no relationship to birthdates. For example:
image

As such, this PR refactors with goals to:

1. Preserve the birth date interface so no backfilling is required from our consumers e.g. CMS and monorepo, etc.
1. Expose the more generic `DateInput`
1. Unplanned -- but it turns out `MM/YYYY` wasn't getting parsed properly by dayjs, but we don't use EDS for those funnel components yet, so it wasn't noticed until I attempted that here -- so this PR fixes and adds those capabilities as well (via dayjs plugins for the most part)

**dayjs plugins**
- Adds https://day.js.org/docs/en/plugin/custom-parse-format to [support things like `MM/YYYY`](https://github.com/iamkun/dayjs/blob/fba621d78461f967d46e752343441e284942f7f8/test/plugin/customParseFormat.test.js#L184)  which don't work without it (parses as Invalid Date). My date validator specs used this odd format so we get some insurance verification there.
- Adds https://day.js.org/docs/en/plugin/utc -- so we can do `dayjs().utc()` similar to how we often do with moment.js. 

___

- [x] Remove duplications from BirthdateInput component and have it just delegate to the new DateInput base component (but don't change BirthdateInput interface so no backfilling required)
- [x] Add unit tests for src/validators/DateInputValidator.js
- [x] Dev QA src/components/BirthdateInput/BirthdateInput.md examples

**Screenshots:**

Manual testing...

Valid date validates:

image

image

Too far in future:

image

Too far in past (1700 is earliest valid in this case):

image

image


src/validators/DateInputValidator.js
```diff
@@ -0,0 +1,103 @@
...
+export const dateMaskByFormat = {
+  'mm/dd/yyyy': [/\d/, /\d/, '/', /\d/, /\d/, '/', /\d/, /\d/, /\d/, /\d/],
+  'mm/yyyy': [/\d/, /\d/, '/', /\d/, /\d/, /\d/, /\d/],
+  'mm/yy': [/\d/, /\d/, '/', /\d/, /\d/],
+}
+
+export const dateRegexByFormat = {
+  'mm/dd/yyyy': /\d\d\/\d\d\/\d\d\d\d/,
+  'mm/yyyy': /\d\d\/\d\d\d\d/,
+  'mm/yy': /\d\d\/\d\d/,
```
keldonia, line 14: nit: This could be cleaned up with the quantifier `/\d{1,4}/` capturing 1-4 digits for example.  If capture groups are included it could also be used for extracting data/formatting the input (if you make the `\` optional be rewrite it... something like `/(?<month>\d{1,2})(?:\/?)(?<year>\d{0,4})/`, which would give named capture groups too, and slash being a special non-capturing group.  The other part of it is that this regex could be tokenized into its components, then the true regex could be composed, eg:
```js
  const month: `(?<month>\d{0,2})`
  const year: `(?<year>\d{0,4})`
  const slash: `(?:\/?)`

  const mmyyyyRegex = new Regex(month + slash + year)
```

roblevintennis: Brian, I really really like you're point here and I totally think it has merit. Let me tell you why I'd like to punt on this though.

This refactor already feels a bit aggressive as I'm completely teasing apart all the BirthdateInput code and delegating. I've captured your sentiments though:
https://app.asana.com/0/1137984670709817/1176569816350953/f

keldonia: That's completely sufficient, as it would definitely be an optimization

jameshawkinsjr:
Something with the validator regex might be off, but on some dates that I type in, both validation error messages pop up (doesn't always happen, but when there's a leading 0, it looks like it might get thrown off: 
Screen Shot 2020-05-19 at 1 11 34 PM


Screen Shot 2020-05-19 at 1 09 00 PM

Screen Shot 2020-05-19 at 1 09 27 PM


Screen Shot 2020-05-19 at 1 09 33 PM

Birthdate input - looks like the leading 0's are being ignored in these cases (it thinks I'm typing in `MM/DD/YY` instead of `MM/DD/YYYY`
Screen Shot 2020-05-19 at 1 12 45 PM

Screen Shot 2020-05-19 at 1 12 14 PM

roblevintennis:
Yeah, it's in dayjs https://github.com/iamkun/dayjs/issues/596 and it's something only with leading zeroes and so it'll get caught upstream and probably quite rare that a user actually has any good reason to type a YYYY in that malformed 0088.

PR #345, opened 2020-06-05
Remove border from NoraButton

Domajno:
**Description:**

- [Asana Task](https://app.asana.com/0/1150068311310825/1178718765502761/f)

All blue buttons in Nora have black outline:
Screenshot 2020-06-05 at 11 57 26 AM
Screenshot 2020-06-05 at 11 56 58 AM

But none of the blue buttons have black border in Figma:
Screenshot 2020-06-05 at 11 55 34 AM
Screenshot 2020-06-05 at 11 55 12 AM

- [x] Followed the checklist at the bottom of the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute) guide?

-

**Screenshots:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.211",
+  "version": "1.1.213",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.1.211",
+  "version": "1.1.213",
```
ankur-ethos, line 3: why do we need to update EDS for this change?

Domajno: Basic style of the nora button comes from EDS. We could override it Nora styles but seems like this may be a better place. I may be wrong though. Not sure who is the best person to ask.

Domajno: Or is the question why we need to bump the version?

PR #381, opened 2021-03-18
Add debounce to buttons

mwwu:
**Description:**

- [Asana Task](https://app.asana.com/0/1116481661798430/1200056096864969)
-

-

**Screenshots:**

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -6,6 +6,8 @@ import debounce from 'lodash.debounce'
 
 import styles from './Button.module.scss'
 
+const DEFAULT_DEBOUNCE_DURATION_MS = 500
+
 /* @getethos/design-system/Button.js
 
    Legend:
@@ -55,7 +57,7 @@ function PrivateButton({
   isSelected, // only used for Button.*.Stateful, aka SelectableHtmlButtonGroup
   arrowIcon,
   backArrowIcon,
-  debounceDurationMs = 500,
+  debounceDurationMs = DEFAULT_DEBOUNCE_DURATION_MS,
   ...rest
 }) {
   // Verify that size, type, and style were valid enum values
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 import useIncludes from '../../hooks/useIncludes.js'
 import useInvalid from '../../hooks/useInvalid.js'
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -1,7 +1,8 @@
-import React from 'react'
+import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import useIncludes from '../../hooks/useIncludes.js'
 import useInvalid from '../../hooks/useInvalid.js'
+import debounce from 'lodash.debounce'
 
 import styles from './Button.module.scss'
 
@@ -54,6 +55,7 @@ function PrivateButton({
   isSelected, // only used for Button.*.Stateful, aka SelectableHtmlButtonGroup
   arrowIcon,
   backArrowIcon,
+  debounceDurationMs = 500,
   ...rest
 }) {
   // Verify that size, type, and style were valid enum values
@@ -93,14 +95,37 @@ function PrivateButton({
     ''
   )
 
+  const [isDebounced, setIsDebounced] = useState(false)
+
+  useEffect(() => {
+    if (isDebounced) {
+      setTimeout(() => setIsDebounced(false), debounceDurationMs)
+    }
+  }, [isDebounced])
+
+  let debouncedOnClick
+  if (typeof onClick === 'function') {
+    debouncedOnClick = debounce(
+      (e) => {
+        onClick(e)
+        setIsDebounced(true)
+      },
+      debounceDurationMs,
+      {
+        leading: true,
+        trailing: false,
+      }
+    )
+  }
+
   return (
     <button
       {...checked}
       className={cssModulesClasses}
-      disabled={disabled}
+      disabled={disabled || isDebounced}
       type={type}
       name={name}
-      onClick={onClick}
+      onClick={debouncedOnClick}
       data-tid={rest['data-tid']}
       role={role}
     >
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -6,7 +6,7 @@ import useDebounce from '../../hooks/useDebounce.js'
 
 import styles from './Button.module.scss'
 
-const DEFAULT_DEBOUNCE_DURATION_MS = 200
+const DEFAULT_DEBOUNCE_DURATION_MS = 100
 
 /* @getethos/design-system/Button.js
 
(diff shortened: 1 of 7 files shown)


src/components/Button/Button.js
```diff
@@ -54,6 +55,7 @@ function PrivateButton({
   isSelected, // only used for Button.*.Stateful, aka SelectableHtmlButtonGroup
   arrowIcon,
   backArrowIcon,
+  debounceDurationMs = 500,
```
mwwu, line 58: will look into DD to figure out the minimum value that makes sense here

src/components/Button/Button.js
```diff
@@ -93,14 +95,37 @@ function PrivateButton({
...
 
+  const [isDebounced, setIsDebounced] = useState(false)
+
+  useEffect(() => {
+    if (isDebounced) {
+      setTimeout(() => setIsDebounced(false), debounceDurationMs)
+    }
+  }, [isDebounced])
+
+  let debouncedOnClick
```
anbright, line 106: What is the behavior if we pass in `undefined` below? Do we want a default no-op?

src/components/Button/Button.js
```diff
@@ -93,14 +95,37 @@ function PrivateButton({
     ''
   )
 
+  const [isDebounced, setIsDebounced] = useState(false)
+
+  useEffect(() => {
+    if (isDebounced) {
+      setTimeout(() => setIsDebounced(false), debounceDurationMs)
+    }
+  }, [isDebounced])
```
anbright, line 104: May make sense to wrap this logic in a hook that can return the `debouncedOnClick` function

src/components/Button/Button.js
```diff
@@ -93,14 +95,37 @@ function PrivateButton({
     ''
   )
 
+  const [isDebounced, setIsDebounced] = useState(false)
```
anbright, line 98: Is it possible to get specs (may be easier if it is a hook)?

iamvictorli (commented):
I wouldn't want the EDS button should have any logic for debouncing, as it's just a front end component. There may be situations where we may not want debouncing on buttons. 
Two solutions I would like:
1. Debouncing logic should be handled when it is used, as in your case at the invoices page. 
2. A Debounced EDS button which wraps the EDS Button with debouncing logic

iamvictorli (commented):
Have you also looked at wrapping `lodash.debounce` wrapping `onClick`? May not need use React state

tkarssli (approved):
Tested locally lgtm

anbright:
> I wouldn't want the EDS button should have any logic for debouncing, as it's just a front end component. There may be situations where we may not want debouncing on buttons.
> Two solutions I would like:
> 
> 1. Debouncing logic should be handled when it is used, as in your case at the invoices page.
> 2. A Debounced EDS button which wraps the EDS Button with debouncing logic

Ideally I think we would want debouncing across the board as the vast majority of use cases should be debounced. I would argue that the button component should define the this functionality so we don't have to configure it in every place that calls this, but may be good to expose a param that can disable this if there is ever a use case for not debouncing.

src/components/Button/Button.js
mwwu: updated, moved into custom hook https://github.com/getethos/ethos-design-system/pull/381/commits/dea7d6c8344728304db8362c2d0c94f69660b956

src/components/Button/Button.js
mwwu: updated specs: https://github.com/getethos/ethos-design-system/pull/381/commits/dea7d6c8344728304db8362c2d0c94f69660b956

src/components/Button/Button.js
mwwu: if value is undefined / some unexpected value, just passing through so it behaves as it would before this change https://github.com/getethos/ethos-design-system/blob/9516b4792afc44792efa5e20a22664767b4607bb/src/hooks/useDebounce.js#L23

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,42 @@
+import { useState, useEffect, useCallback } from 'react'
+import debounce from 'lodash.debounce'
+
+/**
+ * Custom hook to debounce a function for a specified amount of time
+ */
+const useDebounce = (fn, debounceDurationMs) => {
```
cloverich, line 7: You should jsdoc comment these if you can't use typesecript, specifically to explain how `0` disables debouncing. It should probably also be documented at the button level, esp. considering debouncing will be the new default behavior.

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,42 @@
...
+      timer = setTimeout(() => {
+        setIsDebounced(false)
+      }, debounceDurationMs)
+    }
+    return () => {
+      clearTimeout(timer)
+    }
+  }, [isDebounced])
+
+  let debouncedFn = useCallback(
```
cloverich, line 22: why let?

src/hooks/useDebounce.js
```diff
@@ -0,0 +1,42 @@
...
+    typeof fn === 'function' && debounceDurationMs > 0
+      ? debounce(
```
cloverich, line 24: Is this necessary? The outer `setTimeout` + the button's `isDisabled` state already debounces the _button_. i.e.

```ts
const debouncedFn = useCallback(() => {
  if (isDebounced) return;
  setIsDebounced(true);
  fn()
}, [fn, debounceDurationMs])
```

src/hooks/useDebounce.test.js
```diff
@@ -0,0 +1,107 @@
...
+    // invoke function with params
+    const mockParams = { a: 'aaa', b: 'bbb' }
+    act(() => {
+      result.current[1](mockParams) // debouncedFn()
+    })
+    expect(mockFn).toHaveBeenCalledTimes(1)
+    expect(mockFn).toHaveBeenCalledWith(mockParams)
+  })
+
+  test('no errors when target function is not a function', () => {
```
cloverich, line 100: You might explain why this scenario exists. That will let a future refactor decide if they can remove it.

cloverich (approved):
Nice work, nice tests.

src/hooks/useDebounce.js
cloverich: > Have you also looked at wrapping lodash.debounce wrapping onClick? May not need use React state

I see maybe this was your initial implementation, which changed in response to this. I think the answer is you need react state because you want the button's presenrtation to change depending on the debounce state. Also, because the debounced version of the function is encapsulated inside the button (i.e. it returns a wrapped local version), there's no possibility of sharing the debounced version. Technically there is now, but I think that use case is undesireable without a concrete example to ensure it has proper test coverage (i.e. some parent component making a debounced version and sharing with multiple child components).

src/hooks/useDebounce.js
mwwu: updated https://github.com/getethos/ethos-design-system/pull/381/commits/5a3cd0f8eb8ec03f0f4ecaaa45978ae721ac4631

src/hooks/useDebounce.js
mwwu: i discovered it was necessary after implementing unit tests. when I did not use `lodash.debounce` and relied on the state, I could invoke the "debounced" function multiple times before the state/flag was updated. I couldn't click it fast enough to reproduce this by hand, but I thought it would be safer ensure the function was debounced by definition

anbright (approved):
Looks good to me

PR #382, opened 2021-04-15
Change initialization order of variables in select

mwwu:
**Description:**
- unable to `mount` <Select /> because onBlur was being used before it was initialized


jameshawkinsjr (approved):
lgtm

mwwu:
@jameshawkinsjr yeah exactly, its a mystery why it hasn't failed more often tho

PR #401, opened 2021-08-20
button select group supports currentValue being set post inital-render

seanhho:
**Description:**

- [Asana Task]()
-

-

**Screenshots:**

### DIFF
--- a/src/components/ButtonSelectGroup/ButtonSelectGroup.js
+++ b/src/components/ButtonSelectGroup/ButtonSelectGroup.js
@@ -6,6 +6,7 @@ import { OptionButton } from './OptionButton'
 import { InputLabel } from '../InputLabel'
 
 import styles from './ButtonGroup.module.scss'
+import isUndefined from 'lodash/isUndefined'
 
 /**
  **/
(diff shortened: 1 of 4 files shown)


src/components/ButtonSelectGroup/ButtonSelectGroup.js
```diff
@@ -66,6 +67,15 @@ export const ButtonSelectGroup = ({
   // Set up validation hooks
   const [getError, setError, , validate] = useErrorMessage(validator)
 
+  // can set current value if selectedValue is not selected.  This is important for gatsby rendering
+  if (
+    !isUndefined(currentValue) &&
```
ChirashiSushi, line 72: @seanhho what does `isUndefined` check exactly? is it redundant that we are checking both `!isUndefined(currentValue)` and `currentValue !== ''`?

ChirashiSushi (commented):
Are you going to update the EDS version in the next PR?

src/components/ButtonSelectGroup/ButtonSelectGroup.js
ChirashiSushi: Looks like `isUndefined` checks for null or undefined of the value, so my understanding is that `!currentValue` covers all three scenarios: null, undefined, and empty string. Is there anything that I miss here?

seanhho: yes this is true.  Notably in the form.js component it passes in an empty string where as in most situations we pass in nothing/undefined.  I make sure to use the lodash function because javascript has weird issues with undefined that we want to avoid.

seanhho:
@ChirashiSushi I've bumped the version number, I think that's it?  I noticed the readme mentioned updating the version number but it seems the yarn command did it

WesleyDRobinson (approved):
canMergeCurrentPROntoMaster

PR #405, opened 2021-09-03
floating header extension 3

WesleyDRobinson:
**Description:**

Supports Floating Heade


src/components/Select/Select.module.scss
```diff
@@ -13,12 +13,8 @@
     }
 
     .StyledReactSelect__multi-value {
-      margin: 0;
       padding: 2px;
-      margin-left: 0;
-      margin-bottom: 0;
-      margin-top: 2px;
-      margin-right: var(--Space-6);
+      margin: 2px var(--Space-6) 0 0;
```
WesleyDRobinson, line 17: 🏌️

src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -95,45 +95,25 @@ export const TextMaskedInput = (props) => {
...
+    onBlur,
+    name,
+    placeholder: restProps.placeholder,
+    className: getClasses(),
+    disabled: restProps.disabled,
+    placeholderChar,
+    autoComplete,
+  }
+
+  if (typeof mask !== 'function') {
```
WesleyDRobinson, line 113: It was tough to see the difference between the two branches, I hope this helps.

src/components/index.d.ts
```diff
@@ -325,7 +325,17 @@ export declare const CloudinaryImage: {
 }
 export declare const filePath: (publicId: any) => any
 
-export { COLORS } from './Colors'
+export declare const COLORS: {
```
WesleyDRobinson, line 328: my IDE was not finding COLORS with the previous export 🤔 idk why, but this helps.

coleEthos: Is fixing the IDE the only reason you made this change @WesleyDRobinson? 
My concern is anyone updating `Colors` in the future will have to know to manually change the definition here as well. That's probably okay if we're not building in new colors that often but maybe we should add a comment to `Colors.js` explaining the declaration here.

src/components/TextMaskedInput/TextMaskedInput.js
coleEthos: Yeah that's a lot easier to read.

coleEthos (approved):
LGTM - just one question about future maintenance of `Colors`, but not blocking.

coleEthos:
Reminder to re-tag with v1.2.4 @WesleyDRobinson, once done with any changes!

src/components/Select/Select.js
```diff
@@ -23,6 +23,7 @@ export const ReactSelectComponents = components
  */
 export const Select = ({
   className,
+  classOverrides,
```
WesleyDRobinson, line 26: these are meant to pass thru/ override classes for the Select component itself

src/components/Select/Select.js
```diff
@@ -122,12 +115,34 @@ export const Select = ({
...
+  const getClasses = () => {
+    const cName = className || ''
+    const cOverride = classOverrides || ''
+    const compactClass = isCompact ? styles.Compact : ''
+    return `${cName} ${compactClass} ${styles.root} ${cOverride}`
+  }
+
   return (
-    <div className={wrapperClass} data-tid={rest['data-tid']}>
+    <div className={title ? styles.wrapper : ''} data-tid={rest['data-tid']}>
```
WesleyDRobinson, line 126: If it makes things more "readable" I really appreciate the pattern where you do: `const wrapper = <build the styles string>` but here i think it doesn't help too much.

src/components/index.d.ts
WesleyDRobinson: that's a good point. definitely strange behavior on my end, and I assumed for others also. i retried the export from Colors.js (ran prettier) and that helps so far.

PR #420, opened 2021-12-15
Add number range validator

ChirashiSushi:

## Screenshots
https://user-images.githubusercontent.com/84357906/146285367-fa4d515a-42c2-4b02-9a45-d536bd5aaae9.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.15",
+  "version": "1.2.16",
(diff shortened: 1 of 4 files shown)


ChirashiSushi:
I rebased this PR to master, and hence the version for this PR is `v1.2.14` which is the same version as Harry's latest PR

coleEthos (approved):
LGTM - in lieu of #424, could we version this as v1.2.16?

PR #421, opened 2021-12-15
add eyebrow2

harrokun:

- [ ] Read and followed the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute)?
I couldn't get the page to open regardless of my Viscosity setting or lack thereof.
- [x] Type is exported from [index.d.ts](https://github.com/getethos/ethos-design-system/blob/master/src/components/index.d.ts) so Typescript users can consume it?
- [ ] Added a test and ran `yarn test:types`?
Depends on whether what I added counts as a test. The yarn tests were definitely run though.

-

Screen Shot 2021-12-15 at 5 03 56 PM


coleEthos (approved):
LGTM 👁️ 🟫

PR #425, opened 2021-12-17
Slash div migration

ethantran:
To remove warnings from sass output, I had to run this migrator. https://sass-lang.com/documentation/breaking-changes/slash-div

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.16",
+  "version": "1.3.0",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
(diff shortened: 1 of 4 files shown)


coleEthos (approved):
I think this is okay. Couple questions @ethantran 

- Did you test consuming this new version of EDS in anything like main app or CMS? Want to make sure it's not a breaking change in anyway (shouldn't be from my understanding of sass:math). Is the `node-sass` package where we get `sass:math` from?
- Should we also update `ethos-design-system/src/components/Select/Select.module.scss`? I see 3 `line-height` division calculations in that file. 
- Once you're done and before merging, please make sure to also add the git tag for this version 1.3.0. I didn't see on here: https://github.com/getethos/ethos-design-system/tags.

PR #426, opened 2021-12-17
Enable maxLength for NumberInput

ChirashiSushi:

https://user-images.githubusercontent.com/84357906/146617248-4ca54122-a601-4bdf-ba14-80d0966b7f6f.mov

Home Page:
Screen Shot 2021-12-17 at 5 04 20 PM

- now we can use `maxLength` to control the max number of characters allowed in `NumberInput` component

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -803,6 +803,7 @@ export declare const TextMaskedInput: {
     getTouched?: boolean
     placeholderChar?: string
     formChangeHandler?: (value: string, errorValue: string) => void
+    maxLength?: number
   }
   propTypes: {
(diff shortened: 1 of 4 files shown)


src/components/index.d.ts
```diff
@@ -1235,6 +1235,7 @@ export declare const NumberInput: {
     placeholderChar?: string
     guide?: boolean
     keepCharPositions?: boolean
+    maxLength?: number
```
coleEthos, line 1238: Do we also need to add this line to `export declare const TextMaskedInput: {...` further up in this file?

coleEthos (approved):
LGTM

PR #436, opened 2022-03-14
[BUG] Fix Select Component Error Validation

AjaySingh5:

- [X] Read and followed the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute)?
- [X] Bumped the yarn version?

- Saw this originally when running through LGA Prime on physician question, then also realized [CC-377](https://ethoslife.atlassian.net/browse/CC-377) was due to the same error so I wanted to fix it

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.18",
+  "version": "1.2.19",
   "description": "Ethos Design System v1",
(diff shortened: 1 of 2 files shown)


jameshawkinsjr (approved):
Nice Find

WesleyDRobinson (approved):
gorgeous! thanks for finding the root of this issue and fixing!

PR #437, opened 2022-03-15
Bug fix for select component

AjaySingh5:

- reverts what happened in https://github.com/getethos/ethos-design-system/pull/436

- Basically the validators for select didn't like that we used empty string for value [code](https://github.com/getethos/ethos/blob/d9e593a68944a7e94c9159dc1e295fee947825ec/frontend/src/main/InsuranceApp/PreInterview/Questions/validators.ts#L110)
- This fixes that while still fixing the underlying issue of the console error being thrown on initial select Blur

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -77,10 +77,10 @@ export const Select = ({
     let resolvedValues = ''
     // react-select multi select case
     if (Array.isArray(userSelection)) {
-      resolvedValues = userSelection.map((selection) => selection.value)
+      resolvedValues = userSelection?.map((selection) => selection?.value)
     } else {
       // react-select single select case
-      resolvedValues = userSelection.value
+      resolvedValues = userSelection?.value
     }
 
     errorMessage = validate(resolvedValues)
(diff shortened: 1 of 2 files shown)


src/components/Select/Select.js
```diff
@@ -80,10 +77,10 @@ export const Select = ({
     let resolvedValues = ''
     // react-select multi select case
     if (Array.isArray(userSelection)) {
-      resolvedValues = userSelection.map((selection) => selection.value)
+      resolvedValues = userSelection?.map((selection) => selection?.value)
```
WesleyDRobinson, line 80: one last check, do you need to add a `.filter(el => el)` to strip any `undefined` from resolvedValues?

AjaySingh5: Absolutely, [90097bb](https://github.com/getethos/ethos-design-system/pull/437/commits/90097bb5dcaf56e3e7983ec3bdefd8cd4eeb6573)

AjaySingh5: Per slack DMs, [d29d345](https://github.com/getethos/ethos-design-system/pull/437/commits/d29d345adbb44c8e803aad270b867f9dc2094784)

WesleyDRobinson (approved):

PR #438, opened 2022-03-24
allow lazy load disabling on CloudinaryImage

coleEthos:

- [ ] Read and followed the [contributing guidelines](https://eds.ethoslabs.io/#/Guidelines?id=section-contribute)?

-


ChirashiSushi (approved):
LGTM

PR #439, opened 2022-03-30
Enable white background for checkbox

ChirashiSushi:
### Description:
- Enable white background for CheckboxInput by introducing `useWhiteBackground` property


coleEthos (approved):
Lgtm!

PR #443, opened 2022-04-12
Add an icon option for TextInput EmailInput NumberInput

MayUWish:

- [x] Read and followed the [contributing guidelines](https://eds.eks.dev.ethos-int.com/#/Guidelines?id=section-contribute)
- [x] Bumped the yarn version? (=> will do after review)

-

- added options: iconName and iconPrefix to TextInput, EmailInput, NumberInput to allow icon display inside input
- related examples are added at markdown file for TextInput, EmailInput, NumberInput 
## email input screenshots
image

## number input screenshots
image

## text input screenshots
image

### DIFF
--- a/src/components/EmailInput/EmailInput.js
+++ b/src/components/EmailInput/EmailInput.js
@@ -61,10 +61,10 @@ EmailInput.propTypes = {
   initialValue: PropTypes.string,
   autoComplete: PropTypes.string,
   classOverrides: PropTypes.string,
-  iconPrefix: PropTypes.string,
   /** iconPrefix and iconName work together to render icon in input; Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. e.g. `iconPrefix="fas"` is the prefix for solid icons; `iconPrefix="far"` is the prefix for regular icons.*/
-  iconName: PropTypes.string,
+  iconPrefix: PropTypes.string,
   /** iconPrefix and iconName work together to render icon in input; Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. */
+  iconName: PropTypes.string,
 }
 
 EmailInput.defaultProps = {
--- a/src/components/EmailInput/EmailInput.js
+++ b/src/components/EmailInput/EmailInput.js
@@ -62,7 +62,9 @@ EmailInput.propTypes = {
   autoComplete: PropTypes.string,
   classOverrides: PropTypes.string,
   iconPrefix: PropTypes.string,
+  /** iconPrefix and iconName work together to render icon in input; Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. e.g. `iconPrefix="fas"` is the prefix for solid icons; `iconPrefix="far"` is the prefix for regular icons.*/
   iconName: PropTypes.string,
+  /** iconPrefix and iconName work together to render icon in input; Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. */
 }
 
 EmailInput.defaultProps = {
--- a/src/components/EmailInput/EmailInput.js
+++ b/src/components/EmailInput/EmailInput.js
@@ -2,7 +2,7 @@ import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { TextInput } from '../TextInput'
 import EmailFormatValidator from '../../validators/EmailValidator'
-import { valid_icons } from '../../helpers/constants'
+import { VALID_ICONS } from '../../helpers/constants'
 
 export const EmailInput = (props) => {
   const {
@@ -60,8 +60,8 @@ EmailInput.propTypes = {
   initialValue: PropTypes.string,
   autoComplete: PropTypes.string,
   classOverrides: PropTypes.string,
-  /** iconPrefix and iconName work together to render icon in input. Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. Currently allowed icons are defined by valid_icons at src/helpers/constants.js */
-  icon: PropTypes.oneOf(Object.keys(valid_icons)),
+  /** iconPrefix and iconName work together to render icon in input. Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. Currently allowed icons are defined by VALID_ICONS at src/helpers/constants.js */
+  icon: PropTypes.oneOf(Object.keys(VALID_ICONS)),
 }
 
 EmailInput.defaultProps = {
--- a/src/components/EmailInput/EmailInput.js
+++ b/src/components/EmailInput/EmailInput.js
@@ -2,6 +2,7 @@ import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { TextInput } from '../TextInput'
 import EmailFormatValidator from '../../validators/EmailValidator'
+import { valid_icons } from '../../helpers/constants'
 
 export const EmailInput = (props) => {
   const {
@@ -14,8 +15,7 @@ export const EmailInput = (props) => {
     placeholder,
     disabled,
     autoComplete,
-    iconPrefix,
-    iconName,
+    icon,
     ...restProps
   } = props
 
@@ -36,8 +36,7 @@ export const EmailInput = (props) => {
         placeholder={placeholder}
         validator={EmailFormatValidator}
         autoComplete={autoComplete}
-        iconPrefix={iconPrefix}
-        iconName={iconName}
+        icon={icon}
         {...restProps}
       />
     </>
@@ -61,15 +60,11 @@ EmailInput.propTypes = {
   initialValue: PropTypes.string,
   autoComplete: PropTypes.string,
   classOverrides: PropTypes.string,
-  /** iconPrefix and iconName work together to render icon in input; Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. e.g. `iconPrefix="fas"` is the prefix for solid icons; `iconPrefix="far"` is the prefix for regular icons.*/
-  iconPrefix: PropTypes.string,
-  /** iconPrefix and iconName work together to render icon in input; Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. */
-  iconName: PropTypes.string,
+  /** iconPrefix and iconName work together to render icon in input. Please refer to https://fontawesome.com/v5/docs/apis/javascript/import-icons for more information about iconPrefix. Please refer to `fa.js` and https://fontawesome.com for more info about icon's name. Currently allowed icons are defined by valid_icons at src/helpers/constants.js */
+  icon: PropTypes.oneOf(Object.keys(valid_icons)),
 }
 
 EmailInput.defaultProps = {
   labelCopy: 'Email',
   placeholder: '',
-  iconPrefix: '',
-  iconName: '',
 }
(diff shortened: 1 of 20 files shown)


src/components/NumberInput/NumberInput.js
```diff
@@ -88,9 +92,13 @@ NumberInput.propTypes = {
   keepCharPositions: PropTypes.bool,
   autoComplete: PropTypes.string,
   maxLength: PropTypes.number,
+  iconPrefix: PropTypes.string,
```
ChirashiSushi, line 95: @MayUWish could you use `/** */` to add descriptions for what `iconPrefix` and `iconName` do and what values they are expecting for them to work properly? You can find an example just above for `capitalize` property and `mask` property for `NumberInput`.

src/components/NumberInput/NumberInput.md
```diff
@@ -48,3 +48,58 @@ const dollarMaskFunction = createNumberMask({
...
+  data-tid='the-number-input'
+  placeholder='number input'
+  formChangeHandler={formChangeHandlerStub}
+  validator={(n) => {
+    if (n > Number.MAX_SAFE_INTEGER) {
+      return 'Number too large—you have exceeded JavaScript\s powers!!'
+    }
+    return n % 2 === 0 ? '' : 'Must be an even number'
+  }}
+  iconPrefix="fas"
```
ChirashiSushi, line 71: For example, as a user I wouldn't know what `fas` and `far` are and when/how to use them for the icon

src/components/TextInput/TextInput.js
```diff
@@ -140,19 +146,29 @@ function PrivateTextInput({
...
+          name={name}
+          placeholder={rest.placeholder}
+          onChange={onChange}
+          onBlur={onBlur}
+          value={value}
+          data-tid={rest['data-tid']}
+          aria-label={name} // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute
+          autoComplete={autoComplete}
+        />
+        {iconPrefix && iconName && (
```
ChirashiSushi, line 163: @MayUWish you may want to do a more specific case checking here than just confirming `iconPrefix` and `iconName` both exist. E.g. if I provide random values to them that `FontAwesomeIcon` doesn't recognize, we would run into an error. My suggestion is to validate them against specific string input values.

src/components/TextInput/TextInput.module.scss
```diff
@@ -46,3 +46,25 @@
...
+.TextInputIconWrapper{
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: 16px;
+}
+
+.TextInputIcon{
+  width:14px;
+  height:18px;
```
ChirashiSushi, line 63: nit pick. There should be spaces between property name and its value.

src/components/TextInput/TextInput.module.scss
```diff
@@ -46,3 +46,25 @@
...
+  margin: 3px 5px;
+  color: #c2c2c2;
+}
+
+.hasIcon{
+  padding-right:50px;
```
ChirashiSushi, line 69: ditto.

src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -129,9 +135,21 @@ export const TextMaskedInput = (props) => {
         allCaps={allCaps}
         capitalize={capitalize}
       />
-
-      <MaskedInput {...maskedInputProps} />
-
+      <div className={styles.TextInputWrapper}>
+        {iconPrefix && iconName ? (
```
ChirashiSushi, line 139: @MayUWish same concern as above. The validation here should be more specific.

src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -129,9 +135,21 @@ export const TextMaskedInput = (props) => {
...
+          <MaskedInput {...maskedInputProps} className={maskedInputClass} />
+        ) : (
+          <MaskedInput {...maskedInputProps} />
+        )}
+        {iconPrefix && iconName && (
```
ChirashiSushi, line 144: ditto.

src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -163,6 +181,8 @@ TextMaskedInput.PUBLIC_PROPS = {
   labelWeight: PropTypes.string,
   labelClasses: PropTypes.string,
   maxLength: PropTypes.number,
+  iconPrefix: PropTypes.string,
```
ChirashiSushi, line 184: @MayUWish same callout as above. Please add description and expected values for these two properties.

src/components/TextMaskedInput/TextMaskedInput.js
```diff
@@ -9,6 +9,7 @@ import cleanse from '../../helpers/cleanse.js'
 
 import styles from '../TextInput/TextInput.module.scss'
 import errorStyles from '../Errors.module.scss'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 
 export const TextMaskedInput = (props) => {
```
ChirashiSushi, line 14: @MayUWish could we please include some examples using markdowns similar to what you did for `NumberInput` and `EmailInput` ?

ChirashiSushi (changes requested):
Looks good for the most part. Please see my comments for some feedback.

src/components/TextInput/TextInput.module.scss
```diff
@@ -46,3 +46,25 @@
   background-color: var(--GrayStrokeAndDisabled--translucent);
   border-color: transparent;
 }
+
+.TextInputWrapper{
```
ChirashiSushi, line 50: add space between `TextInputWrapper` and `{`

src/components/TextInput/TextInput.module.scss
```diff
@@ -46,3 +46,25 @@
...
+  position: relative;
+}
+
+.TextInputIconWrapper{
```
ChirashiSushi, line 54: ditto

src/components/TextInput/TextInput.module.scss
ChirashiSushi: and space between class name and `{`

src/components/TextInput/TextInput.js
MayUWish: If it is an invalid input, then it won't display icon, instead of an error. Please see the image below ( this example is also added to markdown file as an example) Please let me know if any other action would be needed from me for this one. Thank you very much. @ChirashiSushi 

image

src/components/TextInput/TextInput.module.scss
MayUWish: Fixed. Thank you

src/components/TextInput/TextInput.module.scss
MayUWish: Fixed. Thank you

src/components/TextMaskedInput/TextMaskedInput.js
MayUWish: Same as above.  invalid input => no icon.
image

src/components/TextMaskedInput/TextMaskedInput.js
MayUWish: Same as above. 
image

src/components/TextMaskedInput/TextMaskedInput.js
MayUWish: added. Thank you.

src/components/TextInput/TextInput.module.scss
MayUWish: added. Thank you

src/components/TextInput/TextInput.module.scss
MayUWish: added. Thank you.

src/components/TextInput/TextInput.js
```diff
@@ -123,7 +126,13 @@ function PrivateTextInput({
     } else {
       returnClasses = `${styles.TextInputCommon} ${styles.TextInputStylable}`
     }
-
+    // hasIcon class indicates the text input has icon, to give more padding-right in stylings,to prevent overlapping between icon and long input
+    if (
+      (iconPrefix === 'fas' && iconName === 'lock') ||
+      (iconPrefix === 'far' && iconName === 'eye-slash')
```
coleEthos, line 132: Is there a reason we're limiting users to only using these combinations of icons? In the future if someone wants a different icon, would they have to change this code? I would prefer we don't limit this now, so no changes are needed later to use different icons.

src/components/TextInput/TextInput.js
```diff
@@ -140,19 +149,30 @@ function PrivateTextInput({
...
+          placeholder={rest.placeholder}
+          onChange={onChange}
+          onBlur={onBlur}
+          value={value}
+          data-tid={rest['data-tid']}
+          aria-label={name} // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute
+          autoComplete={autoComplete}
+        />
+        {((iconPrefix === 'fas' && iconName === 'lock') ||
+          (iconPrefix === 'far' && iconName === 'eye-slash')) && (
```
coleEthos, line 167: Again, a user would have to check the source code here to see these are the only valid icon combinations, any reason we can't allow any valid prefix with any valid icon name?

src/components/index.d.ts
```diff
@@ -747,10 +747,14 @@ export declare const EmailInput: {
     labelClasses?: string
     validator?: (value: string) => string
     initialValue?: string
+    iconPrefix?: string
+    iconName?: string
```
coleEthos, line 751: If there are only 3 icons being made available, it seems like this could be a single `icon` parameter to me, with a type that checks the value, if present, is equal to one of [bell, lock, eyeSlash]. I think if we want to limit the user to only these 3 icons, we should be more explicit here so there is a clear source of truth for what icons are valid when type checking, without having to read into the conditional statements in the individual components.

coleEthos: Resolving since we discussed on zoom.

src/components/TextInput/TextInput.js
coleEthos: Resolving since we discussed on zoom.

src/components/TextInput/TextInput.js
coleEthos: Resolving since we discussed on zoom.

src/components/index.d.ts
MayUWish: Updated as discussed. Thank you Cole @coleEthos @ChirashiSushi

src/helpers/constants.js
```diff
@@ -16,3 +16,14 @@ export const codes = {
   DOWN: 40,
   TAB: 9,
 }
+
+export const valid_icons = {
```
ChirashiSushi, line 20: @MayUWish nit pick: please use all capital letters for naming constants and enums. In this case it would be `VALID_ICONS`

ChirashiSushi (approved):
LGTM3
Left a final nit pick for constants naming but otherwise LGTM!

src/components/EmailInput/EmailInput.js
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { TextInput } from '../TextInput'
 import EmailFormatValidator from '../../validators/EmailValidator'
+import { valid_icons } from '../../helpers/constants'
```
coleEthos, line 5: nit: for constants we usually use all caps, VALID_ICONS

coleEthos: ah i think you fixed this just as I typed it!

src/helpers/constants.js
MayUWish: Hi @ChirashiSushi  fixed! Thank you!

coleEthos (approved):
LGTM!

PR #445, opened 2022-05-19
follow required props of CloudinaryImage component, make props optional

joshuarule:
Simple update to mark `<CloudinaryImage >` typescript definition props as optional to follow the components required props.

https://github.com/getethos/ethos-design-system/blob/master/src/components/Images/Images.js#L38


AjaySingh5 (approved):
let 'er rip

PR #447, opened 2022-06-15
mxJames / Make all EDS colors available in TypeBase

jwhubert91:

- [Ethos Monorepo trunk](https://github.com/getethos/ethos)

Screen Shot 2022-06-16 at 10 57 50 AM


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.25",
```
jwhubert91, line 3: The tag 1.2.26 could not be reused, and I saw that the repo in prod had the tag `1.2.26` (link: [EDS Release Tags history](https://github.com/getethos/ethos-design-system/releases)). It seemed to me the version should be bumped to match the tag?

src/components/Body2.md
```diff
@@ -19,11 +19,17 @@
...
+  <Body2.Regular400 color={COLORS.BRAND_DUCKEGG}>
+    Brand Duckegg color
+  </Body2.Regular400>
   <Body2.Regular400 color={COLORS.BRAND_FOREST}>
     Brand Forest color
   </Body2.Regular400>
   <Body2.Regular400 color={COLORS.BRAND_SALAMANDER}>
     Brand Salamander color
   </Body2.Regular400>
+  <Body2.Regular400 color={COLORS.GRAY_SECONDARY}>
```
jwhubert91, line 31: Should I create examples for every available color? If we end up doing this for other components too it could get pretty repetitive.

package.json
jwhubert91: I confirmed from the diff of the last release that the tag had been updated to 1.2.26 but not the version. This seems like a mistake.
Link: https://github.com/getethos/ethos-design-system/pull/445/files

src/components/Body2.md
joshuarule: No need.

jameshawkinsjr (approved):

PR #449, opened 2022-06-27
Jrd/attempt to fix eds deploy

joshuarule:
[the problem](https://github.blog/changelog/2021-12-10-github-actions-github-hosted-runners-now-run-node-js-16-by-default/)

Set the node version of the deploy


.github/workflows/deploy.yml
```diff
@@ -27,6 +27,11 @@ jobs:
           @fortawesome:registry=https://npm.fontawesome.com/
           //npm.fontawesome.com/:_authToken=${{ secrets.FONTAWESOME_TOKEN_GITHUB }}
           EOF
+      - uses: actions/checkout@v3
```
t1mwillis, line 30: OK, digging a little deeper, I think you can simply add the on line 13 so we're not using the same action twice
```
- name: Use Node.js
  uses: actions/setup-node@v3
  with:
  node-version: '12.x'
```

joshuarule: 🧠

t1mwillis (approved):

PR #450, opened 2022-06-28
Fix Modal X button in Safari and Add style overrides for `NumberInput` Component

ChirashiSushi:

https://user-images.githubusercontent.com/84357906/176282490-dbb4226c-3827-45a5-8830-d5d98410c48a.mov

### DIFF
--- a/src/components/NumberInput/NumberInput.js
+++ b/src/components/NumberInput/NumberInput.js
@@ -35,6 +35,7 @@ export const NumberInput = (props) => {
     autoComplete,
     maxLength,
     icon,
+    classOverrides,
     ...restProps
   } = props
 
@@ -64,6 +65,7 @@ export const NumberInput = (props) => {
         autoComplete={autoComplete}
         maxLength={maxLength}
         icon={icon}
+        classOverrides={classOverrides}
       />
     </>
   )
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.module.scss
```diff
@@ -140,8 +139,8 @@
   }
 
   .closeButtonCircle {
-    width: var(--Space-12);
-    height: var(--Space-12);
+    width: var(--Space-32);
```
ChirashiSushi, line 142: This was incorrectly set in the first place but somehow Chrome ignored the width and height limit and showed the full X button anyway. Safari doesn't want to cope with this error.

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -120,8 +120,7 @@
   width: calc(100vw - var(--Space-24));
   padding: var(--Space-64) var(--Space-24) var(--Space-80) var(--Space-24);
   max-height: 95vh;
-  overflow-y: scroll;
-
+  overflow-y: auto;
```
ChirashiSushi, line 123: we don't need to show the vertical scroll all the time on mobile

src/components/NumberInput/NumberInput.js
```diff
@@ -35,6 +35,9 @@ export const NumberInput = (props) => {
     autoComplete,
     maxLength,
     icon,
+    classOverrides,
```
ChirashiSushi, line 38: add these capabilities to `NumberInput` so we don't have to use custom CSS to overwrite. A follow up PR in CMS will use these properties for the `Age` question in EW.

WesleyDRobinson (approved):
lgtm 🚢

PR #451, opened 2022-06-28
Update README.md

ChirashiSushi:

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,14 +99,13 @@ Each time you need to consume recent EDS changes elsewhere, you need to bump the
 
 1. To bump the package.json version, run `yarn version` in your branch, This will:
 
-- Prompt you for the new version number
-- Update `package.json` with the version number
+- Prompt you for the new version number (yarn version also creates a new git tag using this version number)
+- Updates `package.json` with the version number
 - Commit the change in a new commit
 - Get your PR approved in your branch and merge it to master
 
-2. To create and push a git tag:
+2. To push a git tag:
 
-- Switch to `master` branch and pull in your latest change
 - Run `git push origin v{new version number}` this will create the new tag with the version number in Github
 - Draft a new release in Github using the new tag
 
--- a/README.md
+++ b/README.md
@@ -102,11 +102,13 @@ Each time you need to consume recent EDS changes elsewhere, you need to bump the
 - Prompt you for the new version number
 - Update `package.json` with the version number
 - Commit the change in a new commit
+- Get your PR approved in your branch and merge it to master
 
-2. To create and push a git tag, do one of these:
+2. To create and push a git tag:
 
-- Run `git tag v1.2.3 && git push origin v1.2.3` on your branch.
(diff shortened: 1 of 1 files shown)


WesleyDRobinson (approved):
oops! I did not follow this update in my most recent change.
I tagged my PR branch and pushed that.

I'll follow this procedure next time.

README.md
```diff
@@ -102,11 +102,13 @@ Each time you need to consume recent EDS changes elsewhere, you need to bump the
 - Prompt you for the new version number
 - Update `package.json` with the version number
 - Commit the change in a new commit
+- Get your PR approved in your branch and merge it to master
 
-2. To create and push a git tag, do one of these:
+2. To create and push a git tag:
```
coleEthos, line 107: This should just say `To push a git tag:`  


[IIUC docs](https://classic.yarnpkg.com/en/docs/cli/version#toc-git-tags) the tag is being created when you run `yarn version`.
> If you run yarn version within a Git repository an [annotated Git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) will be created by default following the format v0.0.0.

README.md
```diff
@@ -102,11 +102,13 @@ Each time you need to consume recent EDS changes elsewhere, you need to bump the
...
 
-- Run `git tag v1.2.3 && git push origin v1.2.3` on your branch.
-- Create a release `v1.2.3` on GitHub, if it's already in trunk.
+- Switch to `master` branch and pull in your latest change
+- Run `git push origin v{new version number}` this will create the new tag with the version number in Github
+- Draft a new release in Github using the new tag
```
coleEthos, line 111: What's the value in doing these releases? They aren't technically required, only the package.json and git tag combined are needed by consumers of EDS.

README.md
```diff
@@ -102,11 +102,13 @@ Each time you need to consume recent EDS changes elsewhere, you need to bump the
...
 - Update `package.json` with the version number
 - Commit the change in a new commit
+- Get your PR approved in your branch and merge it to master
 
-2. To create and push a git tag, do one of these:
+2. To create and push a git tag:
 
-- Run `git tag v1.2.3 && git push origin v1.2.3` on your branch.
-- Create a release `v1.2.3` on GitHub, if it's already in trunk.
+- Switch to `master` branch and pull in your latest change
```
coleEthos, line 109: I don't think this step is needed. As long as the last commit in your PR before it was merged, is the one where you ran `yarn version`, the tag should be on that commit. Pushing the tag doesn't require you be on any specific branch, does it?

README.md
coleEthos: I know we already had it in the readme, not sure why we started doing it is all.

ChirashiSushi: I suppose creating a new release with the most recent changes since the previous release is more for bookkeeping?

coleEthos: Yeah fair enough. I guess the tags just seem to do that job to some extent as well already.

coleEthos (approved):
📖 ✅

PR #452, opened 2022-06-30
Jrd/update checkbox facade

joshuarule:

- To allow for facade functions to get access to the `isChecked` value of the checkbox
- allows passing unique ids to checkbox (previously was forcing `id` to be `name`


src/components/CheckboxInput2/CheckboxGroup.js
```diff
@@ -0,0 +1,130 @@
...
+export default function CheckboxGroup({
+  className = '',
+  options,
+  onChange,
+  formChangeHandler,
+  allCaps = true,
+  capitalize,
+  name,
+  initialValue = [],
+  // currentError,
```
jameshawkinsjr, line 15: Note to remove commented out lines

jameshawkinsjr (approved):
lgtm - only request would be to clean up the commented lines in `CheckboxGroup.js`

PR #453, opened 2022-06-30
Tooltip close button position

ChirashiSushi:

Screen Shot 2022-06-30 at 2 15 08 PM

### DIFF
--- a/src/components/Tooltip/Tooltip.module.scss
+++ b/src/components/Tooltip/Tooltip.module.scss
@@ -142,7 +142,7 @@
     width: var(--Space-32);
     height: var(--Space-32);
     top: var(--Space-12);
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.module.scss
```diff
@@ -142,7 +142,7 @@
     width: var(--Space-32);
     height: var(--Space-32);
     top: var(--Space-12);
-    right: var(--Space-32);
+    right: var(--Space-12);
```
ChirashiSushi, line 145: get close button more centered between top and right edges

ChirashiSushi:
I'm tempting to not release a new version of EDS for a trivial fix like this. This can go out as part of the next main feature changes.

PR #458, opened 2022-08-09
Add un-clickable backdrop to Modal

ChirashiSushi:

## **Screenshot**

https://user-images.githubusercontent.com/84357906/183788908-f17de6df-2595-4b90-aabb-b531acf1eed8.mov

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -80,7 +80,7 @@ export const Modal = ({
         isOpen ? styles.ModalBackdropActive : styles.ModalBackdrop
       )
     }, 300)
-  })
+  }, [])
 
   /**
    * Handler will set the the modal onDismiss to `false` when the escape key is
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React, { useRef, useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { Portal } from '../Portal'
 import styles from './Modal.module.scss'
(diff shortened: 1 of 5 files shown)


src/components/Modal/Modal.js
```diff
@@ -72,6 +72,15 @@ export const Modal = ({
...
 }) => {
   const classes = isOpen ? styles.ModalWrapperActive : styles.ModalWrapper
+  const [backDropClass, setBackDropClass] = useState(styles.ModalBackdrop)
+
+  useEffect(() => {
+    setTimeout(() => {
+      setBackDropClass(
+        isOpen ? styles.ModalBackdropActive : styles.ModalBackdrop
+      )
+    }, 1000)
```
coleEthos, line 82: Why are we waiting a full second here?

src/components/Modal/Modal.js
```diff
@@ -72,6 +72,15 @@ export const Modal = ({
...
+  })
```
coleEthos, line 83: Add empty dependency array?

coleEthos (commented):
This seems like an okay approach. Couple minor questions above. Going to pull the branch and do a little bit of local QA before approving.

coleEthos (changes requested):
I think we need to lower that timeout value from 1 second to be much less or find a way that doesn't rely on a timeout. If a user closes the modal and tries to click something right away, the backdrop blocks them from clicking. I was able to reproduce this locally and verify by increasing to 5000ms in the timeout. 1000 is less noticeable but still could be an issue.

src/components/Variables.scss
```diff
@@ -5,5 +5,6 @@
 $z-index-navbar: 1;
 $z-index-middle: 2;
 $z-index-modal: 3;
+$z-index-backdrop: 1000;
```
WesleyDRobinson, line 8: 🦖 
⛰️

src/components/Modal/Modal.js
ChirashiSushi: Adding empty dependency array here actually makes `useEffect` not working for some reason.

coleEthos: I'm not sure why I suggested that. I'm realizing now this probably needs to run every render not just on the first one.

coleEthos (approved):
LGTM, thanks for lowering the timeout. We should be sure to double check the components that consumer this when we update CMS & main app to check for any special cases where this might cause issues, but based on testing locally in EDS alone, it seems to work fine!

PR #459, opened 2022-08-17
Update Package.json version number

ChirashiSushi:


coleEthos (commented):
I don't see a tag here for v1.2.34 yet. https://github.com/getethos/ethos-design-system/tags
Can you add that to https://github.com/getethos/ethos-design-system/pull/459/files before I approve?

coleEthos (approved):
🔢

PR #460, opened 2022-10-26
Ethan/nav bar analytics

ethanrjoseph:
## [Jira Task]([GC-1312](https://ethoslife.atlassian.net/browse/GC-1312))

Adding analytics tracking to all of the menu components in UniversalNavbarExpanded, ensuring tracking works for desktop and mobile

Currently this does not up the version - planning to run yarn version after it has been reviewed to ensure the tag is on the latest commit

-

### DIFF
--- a/src/components/UniversalNavbarExpanded/MobileNav/MobileNav.js
+++ b/src/components/UniversalNavbarExpanded/MobileNav/MobileNav.js
@@ -164,7 +164,7 @@ const MobileNav = ({
           currentPageCondition={showMobileMenu}
           LinkComponent={LinkComponent}
           trackingFunction={itemTrackingFunction}
-          itemLabel={'AnimatedLogo'}
+          itemLabel={'Logo'}
         >
           <FancyAnimatedLogo />
         </NavLink>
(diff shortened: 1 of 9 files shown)


coleEthos (commented):
@ethanrjoseph do you think we could include a demo-able function in the sample markdown file for this feature? For example in `ethos-design-system/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.md` would we be able to pass in the attribute like so for testing? 
```
<UniversalNavbarExpanded logoHref={'/#/Components/UniversalNavbarExpanded'} links={CMS_LINKS} trackItemClick={(properties)=> {console.log(`trackItemClick props=${properties}`)}}/>
```

Not sure if that will actually work but would be helpful for QA. No worries if it's a bit too tricky to get working with the props.

src/components/UniversalNavbarExpanded/MobileNav/MobileNav.js
```diff
@@ -157,6 +163,8 @@ const MobileNav = ({
           currentPageFunction={(e) => toggleHamburger(e)}
           currentPageCondition={showMobileMenu}
           LinkComponent={LinkComponent}
+          trackingFunction={itemTrackingFunction}
+          itemLabel={'AnimatedLogo'}
```
coleEthos, line 167: This could probably just say logo rather than include animated?

coleEthos: Or maybe mobile logo?

coleEthos (approved):
LGTM - excited to see it in action! Will be a bit easier to deeply QA then (on CMS) and make sure we are covering all the different clickable elements across mobile and desktop.

PR #468, opened 2022-12-09
add trackingFunction to tooltip

ethanrjoseph:
## [Jira Task](https://ethoslife.atlassian.net/browse/GC-1537)

Adding the capability to track views on the Tooltip


coleEthos (approved):
lgtm!

PR #476, opened 2022-12-13
Only fire trackingFunction in ToolTip once

coleEthos:
Follow up to #468 . Prevent excessive tracking events from firing on subsequent hovers or mobile modal interaction.


ethanrjoseph (approved):
lgtm

PR #480, opened 2022-12-20
[GC-1587] Update Cloudinary Package

t1mwillis:

- Mostly just read the documentation and made sure the images still looked good 😄

### DIFF
--- a/src/components/index.d.ts
+++ b/src/components/index.d.ts
@@ -383,7 +383,7 @@ export declare const filePath: (publicId: any) => any
 export declare const preloadImageData: ({
   crop, publicId, height, width,
 }: {
-  crop: keyof typeof CloudinaryImage.CROP_METHODS,
+  crop: 'fit' | 'fill' | 'crop',
   /* the cloudinary image resource URL */
   publicId: string,
   height: number[],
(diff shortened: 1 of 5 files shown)


coleEthos:
@t1mwillis do you think it would be worth getting a deploy preview up on CMS to QA this before we merge? Saves having to go back and version again if fixes are needed. 

You can open a new PR in CMS and in package.json just change `#v1.2.43` to `#tim/gc-1587-cloudinary` and run `yarn` to install this branch. 

Line to edit in package.json is: `"ethos-design-system": "github:getethos/ethos-design-system#v1.2.43",`

I'd feel better about doing a full QA there over just QAing with the EDS demo page.

coleEthos (approved):
LGTM, nice migration work Tim!

src/components/index.d.ts
```diff
@@ -383,7 +383,7 @@ export declare const filePath: (publicId: any) => any
 export declare const preloadImageData: ({
   crop, publicId, height, width,
 }: {
-  crop: keyof typeof CloudinaryImage.CROP_METHODS,
+  crop: string,
```
t1mwillis, line 386: To the person in the future that comes here, I was trying to do the right thing, but our current TS implementation doesn't support the `as const` keyword. `keyof typeof` here actually gives the type of the keys which is `FIT | FILL | CROP` and not the values, `fit | fill | crop`. `string` is just slightly better than what it was -- `any` 😄

PR #483, opened 2023-01-04
[GC-1587] Remove data-srcset when not lazyloading

t1mwillis:

- I would like for <source> elements to have a `srcset` attribute when they are not lazy loaded, this way we get some of the benefits of cloudinary image, webp, etc.
- There will be a separate CMS PR where we leverage this.

[GC-1587]: https://ethoslife.atlassian.net/browse/GC-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Images/Images.js
+++ b/src/components/Images/Images.js
@@ -85,14 +85,26 @@ export const CloudinaryImage = ({
   }
 
   const buildImageTag = (srcSet) => {
-    const srcSetString = srcSet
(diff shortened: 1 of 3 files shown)


coleEthos (approved):
LGTM!

PR #485, opened 2023-01-10
Lazy Load Cloudinary Resource

ChirashiSushi:

**lazy load in live**

https://user-images.githubusercontent.com/84357906/211723439-a2a7b665-68be-47ae-972c-7ec8d13532d4.mov

**current**
- During the initial page load we are sending `67` requests for `564kb` resource to be transferred 
current

**post-change**
-  During the initial page load we are sending `14` total requests for `201kb` resource to be transferred
lazyload

### DIFF
--- a/src/components/Images/Images.js
+++ b/src/components/Images/Images.js
@@ -40,6 +40,7 @@ export const CloudinaryImage = ({
   width,
   height,
   crop,
+  lazyLoad,
   fetchpriority,
   ...rest
 }) => {
@@ -89,17 +90,19 @@ export const CloudinaryImage = ({
       ...baseImageSettings,
     })
 
-    return (
-      <img
-        key={publicId}
-        className={[styles.Image, ...imageClasses].join(' ')}
-        src={srcString}
-        srcSet={srcSetString}
-        alt={alt}
-        fetchpriority={fetchpriority}
-        loading="lazy"
-      />
-    )
+    const imgProps = {
+      key: publicId,
+      className: [styles.Image, ...imageClasses].join(' '),
+      src: srcString,
+      srcSet: srcSetString,
+      fetchpriority: fetchpriority,
+    }
+
+    if (lazyLoad) {
+      Object.assign(imgProps, { loading: 'lazy' })
+    }
+
+    return <img alt={alt} {...imgProps} />
   }
 
   const buildTags = () => {
@@ -170,16 +173,18 @@ export const CloudinaryImage = ({
       baseSvgSettings
     )
 
-    return (
-      <img
-        key={publicId}
-        src={svgUrl}
-        className={[styles.Svg, ...imageClasses].join(' ')}
-        alt={alt}
-        fetchpriority={fetchpriority}
-        loading="lazy"
-      />
-    )
+    const imgProps = {
+      key: publicId,
+      src: svgUrl,
+      className: [styles.Svg, ...imageClasses].join(' '),
+      fetchpriority: fetchpriority,
+    }
+
+    if (lazyLoad) {
+      Object.assign(imgProps, { loading: 'lazy' })
+    }
+
+    return <img alt={alt} {...imgProps} />
   }
 
   // Serve a simpler version if resource is SVG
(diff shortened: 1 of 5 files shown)


src/components/Images/Images.js
```diff
@@ -2,13 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import useRequired from '../../hooks/useRequired.js'
 import useInvalid from '../../hooks/useInvalid.js'
-
 import { createCloudinaryLegacyURL } from '@cloudinary/url-gen'
-
-import { v4 as uuidv4 } from 'uuid'
-// eslint-disable-next-line
-import lazysizes from 'lazysizes'
```
ChirashiSushi, line 10: we no longer rely on `lazysizes` since now `` tag natively supports `loading="lazy"`

src/components/Images/Images.js
```diff
@@ -2,13 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import useRequired from '../../hooks/useRequired.js'
 import useInvalid from '../../hooks/useInvalid.js'
-
 import { createCloudinaryLegacyURL } from '@cloudinary/url-gen'
-
-import { v4 as uuidv4 } from 'uuid'
```
ChirashiSushi, line 8: during testing I realized that we cannot generate an unique uuid as the key for every `img` element on the page, as it causes React to re-request the same image during scrolling of the page since they have different keys.

src/components/Images/Images.js
```diff
@@ -46,7 +40,6 @@ export const CloudinaryImage = ({
   width,
   height,
   crop,
-  lazyLoad,
```
ChirashiSushi, line 49: we should no longer use `lazyload` flag for `Cloudinary` image. I will have a follow-up PR in CMS to remove the use of this flag.

src/components/Images/Images.js
```diff
@@ -162,19 +148,7 @@ export const CloudinaryImage = ({
         imageSettings
       )
 
-      const urlWithChainedLqipTransformation = createCloudinaryLegacyURL(
```
ChirashiSushi, line 165: After discussing with @t1mwillis , we believe that not using LQIP at all reduces the size of the img resource requested substantially and is a worthy tradeoff.

src/components/Images/Images.js
```diff
@@ -198,10 +172,12 @@ export const CloudinaryImage = ({
 
     return (
       <img
-        data-src={svgUrl}
+        key={publicId}
+        src={svgUrl}
         className={[styles.Svg, ...imageClasses].join(' ')}
         alt={alt}
         fetchpriority={fetchpriority}
+        loading="lazy"
```
ChirashiSushi, line 180: natively supported by all browsers

t1mwillis: Same here, let's respect the prop.

t1mwillis (changes requested):
Please continue to respect the `lazyLoad` prop.

src/components/Images/Images.js
WesleyDRobinson: furthermore, there are shorter unique values we could use for making component keys unique per build. We( don't need the level of uniqueness that uuidv4 provides, which is:

> For there to be a one in a billion chance of duplication, 103 trillion version 4 IDs must be generated.

src/components/Images/Images.js
WesleyDRobinson: gotta give props where props are due, respect.

src/components/Images/Images.js
coleEthos: @ChirashiSushi Firefox only lists partial support and Safari says it requires an experimental feature to be enabled. I don't think this would qualify as native support.

coleEthos: I left a comment here originally but misread the spec, I see it's only iframes that don't have full support. LGTM!

coleEthos:
+1 to keeping the lazyload prop, otherwise looking good!

t1mwillis (approved):
LGTM, nice work

coleEthos (approved):

PR #486, opened 2023-01-10
[GC-1625] New Cypress color & button style used in homepage color palette experiment

coleEthos:
## [Jira Task](https://ethoslife.atlassian.net/browse/GC-1625)
Supports new buttons used in multiple modules for the new homepage rebrand design experiment.
Easier to add the new color here than to write custom CSS to replicate button styles in CMS. 
The color variable will also be required for some custom CSS in CMS for this experiment as well.

https://github.com/getethos/cms/pull/6876

Screen Shot 2023-01-10 at 2 31 55 PM
Screen Shot 2023-01-10 at 2 28 24 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.45",
+  "version": "1.3.0",
   "description": "Ethos Design System v1",
   "main": "./src/components/index.js",
   "types": "./src/components/index.d.ts",
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -199,6 +199,7 @@ PrivateButton.STYLES = {
   BLACK: 'Black',
   BLACK_OUTLINE: 'BlackOutline',
   SALAMANDER: 'Salamander',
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.2.45",
+  "version": "1.3.0",
```
coleEthos, line 3: New year, figured why not bump to 1.3? We should really define a release process around when we increment based on some sort of major/minor/revision logic. 

For now v1.3 is just marking the start of 2023 our potential rebranded color palette!

EvitaAFD (approved):
LGTM! Thank you!

package.json
WesleyDRobinson: i agree that a new color pallete is an excellent reason to push a "minor" upgrade!

2.0.0 should mean "some things from 1.x.x are no longer available." An example might be any of the X2 migrations being finalized, so Type2 is no longer available after the upgrade to `2.0.0`

PR #489, opened 2023-01-28
Make tooltip device independent interactable

EvitaAFD:
## [GC-1078](https://ethoslife.atlassian.net/browse/GC-1078?atlOrigin=eyJpIjoiZWRjOGY1ZWRiNmI0NDg4ODg4ZWU3MjYxMmFmZTlmYjciLCJwIjoiaiJ9)
[GC-1076](https://ethoslife.atlassian.net/browse/GC-1076?atlOrigin=eyJpIjoiZGZmMzczNmRhZmNlNDkxM2I2ZmNjZjRiNWZhNjBlYjMiLCJwIjoiaiJ9)

- https://ethoslife.atlassian.net/browse/GC-1076?atlOrigin=eyJpIjoiZGZmMzczNmRhZmNlNDkxM2I2ZmNjZjRiNWZhNjBlYjMiLCJwIjoiaiJ9
- https://ethoslife.atlassian.net/browse/GC-1078?atlOrigin=eyJpIjoiZWRjOGY1ZWRiNmI0NDg4ODg4ZWU3MjYxMmFmZTlmYjciLCJwIjoiaiJ9

Screenshot 2023-01-31 at 11 48 52 AM

[GC-1078]: https://ethoslife.atlassian.net/browse/GC-1078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GC-1076]: https://ethoslife.atlassian.net/browse/GC-1076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -111,16 +111,17 @@ export const Tooltip = ({
         <Reference>
           {({ ref }) =>
             children ? (
+              <div ref={ref} {...referenceProps}>
+                {children}
+              </div>
+            ) : (
               <div
+                className={styles.icon}
                 ref={ref}
                 onFocus={debouncedSetTooltipVisibility(true)}
                 onBlur={debouncedSetTooltipVisibility(false)}
                 {...referenceProps}
               >
-                {children}
-              </div>
-            ) : (
-              <div className={styles.icon} ref={ref} {...referenceProps}>
                 {tooltipVisible ? Tooltip.SVGS.iconHover : Tooltip.SVGS.icon}
               </div>
             )
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -89,6 +89,8 @@ export const Tooltip = ({
     onMouseOver: () => debouncedSetTooltipVisibility(true),
     onMouseOut: () => debouncedSetTooltipVisibility(false),
     onClick: () => setModalVisibility(true),
+    onFocus: () => debouncedSetTooltipVisibility(true),
+    onBlur: () => debouncedSetTooltipVisibility(false),
   }
 
   const popperProps = {
--- a/src/components/Tooltip/Tooltip.js
+++ b/src/components/Tooltip/Tooltip.js
@@ -14,7 +14,7 @@ import styles from './Tooltip.module.scss'
 const BREAKPOINTS = Media.BREAKPOINTS
 const HEADER_ID = 'mobile-modal-heading'
 const DESC_ID = 'mobile-modal-description'
-const ACCESSIBILTY_ID = 'info-tooltip'
+const ACCESSIBILITY_ID = 'info-tooltip'
 
 export const Tooltip = ({
   label,
@@ -122,7 +122,7 @@ export const Tooltip = ({
                 role="button"
                 className={styles.icon}
                 ref={ref}
-                ariaLabelledBy={ACCESSIBILTY_ID}
+                ariaLabelledBy={ACCESSIBILITY_ID}
                 tabIndex={0}
                 {...referenceProps}
               >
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.js
```diff
@@ -115,7 +117,14 @@ export const Tooltip = ({
                 {children}
               </div>
             ) : (
-              <div className={styles.icon} ref={ref} {...referenceProps}>
+              <div
+                role="button"
```
EvitaAFD, line 121: The role is required for this to be seen as interactive and therefore allow the user to tab through the page and open the tooltip, divs shouldn't have a tabIndex on them unless you specify that they are in fact an interactive element, in this case the div is acting like a button, so this is the appropriate role.

src/components/Tooltip/Tooltip.js
```diff
@@ -115,7 +117,14 @@ export const Tooltip = ({
...
+                className={styles.icon}
+                ref={ref}
+                aria-labelledby="info-tooltip"
+                tabIndex={0}
```
EvitaAFD, line 125: The tabIndex paired with the onFocus and onBlur are what allow users who rely solely on the keyboard to navigate the screen to be able to interact with this element

src/components/Tooltip/Tooltip.js
```diff
@@ -115,7 +117,14 @@ export const Tooltip = ({
                 {children}
               </div>
             ) : (
-              <div className={styles.icon} ref={ref} {...referenceProps}>
+              <div
+                role="button"
+                className={styles.icon}
+                ref={ref}
+                aria-labelledby="info-tooltip"
```
t1mwillis, line 124:
```suggestion
                ariaLabelledBy="info-tooltip"
```

t1mwillis:
Ultimately this looks good to me. A couple questions:
* Is there anyway to "check" our work here? Can we run an audit on the deploy preview?
* A little higher in the file, there are some `HEADER_ID` & `DESC_ID` variables, does it make sense to use those for our `labeledBy` value?

EvitaAFD:
> Ultimately this looks good to me. A couple questions:
> 
> * Is there anyway to "check" our work here? Can we run an audit on the deploy preview?
> * A little higher in the file, there are some `HEADER_ID` & `DESC_ID` variables, does it make sense to use those for our `labeledBy` value?

I tested it in the styleguide, you can see the screenshot in the PR where the focus box is on the 'i' icon- you can try to test this navigating your webpage using the tab key, you can enable a focus box in your accessibility settings to help make it more obvious that an element is selected. I am not familiar with those variables - are they specifically for accessibility?

t1mwillis:
OK -- I'll have to do some accessibility testing reading!

Those variables are are used by the "mobile" version of the tooltip, the modal. Check out line 61 in the file. Mostly just asking if it makes sense to use them for the icon, or at least declare them similarly as constants instead of the magic string `info-tooltip`. Up to you!

t1mwillis (approved):

t1mwillis (changes requested):
Actually, let's commit that suggestion so we don't run into linter issues!

src/components/Tooltip/Tooltip.js
```diff
@@ -14,6 +14,7 @@ import styles from './Tooltip.module.scss'
 const BREAKPOINTS = Media.BREAKPOINTS
 const HEADER_ID = 'mobile-modal-heading'
 const DESC_ID = 'mobile-modal-description'
+const ACCESSIBILTY_ID = 'info-tooltip'
```
t1mwillis, line 17:
```suggestion
const ACCESSIBILITY_ID = 'info-tooltip'
```

t1mwillis: Then let's get this in!

t1mwillis (approved):

PR #492, opened 2023-02-14
[GC-1883]Transparent/multicolor navbar with animated background

hangngo21:

-

### DIFF
--- a/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
+++ b/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
@@ -280,7 +280,10 @@ UniversalNavbarExpanded.propTypes = {
     href: PropTypes.string,
     title: PropTypes.string,
   }),
+  /** Navigation bar color*/
   navBarColor: PropTypes.string,
+  /** Add animation to nav bar*/
+  animateNavbar: PropTypes.bool,
 }
 
 UniversalNavbarExpanded.defaultProps = {
--- a/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
+++ b/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
@@ -121,6 +121,7 @@ const UniversalNavbarExpanded = ({
       <div className={styles.navbar}>
         <Layout.ScrollDetector>
           <MobileNav
+            navbarColor={navbarColor}
             ctaButtonStyle={ctaButtonStyle}
             extraClass={'isFixedCta'}
             logoHref={logoHref}
--- a/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
+++ b/src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
@@ -55,6 +55,7 @@ const UniversalNavbarExpanded = ({
   links,
   estimateExperiment,
   singleCta = {},
+  navbarColor,
 }) => {
   let BELOW_ACCORDION_LINKS = [links.CTA]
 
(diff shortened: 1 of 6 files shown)


src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
```diff
@@ -89,6 +88,18 @@ $laptop-range-small-end: 929px;
   z-index: $z-index-navbar;
 }
 
+.bgMoss {
+  background-color: var(--BrandMoss);
+  color: var(--BrandForest);
+}
```
coleEthos, line 94: Would it be difficult to have this just be a transparent background, so that whatever module we place behind the navbar automatically sets the background color? 

It would also allow us to fade in the white background on scroll using the alpha channel.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.js
```diff
@@ -267,6 +285,10 @@ UniversalNavbarExpanded.propTypes = {
     href: PropTypes.string,
     title: PropTypes.string,
   }),
+  /** Add color to nav bar*/
+  navbarColor: PropTypes.string,
```
coleEthos, line 289: I think if we could go with just a transparent background to start, when `animateNavbar` is set to `true`, then we could do away with this `navbarColor` prop. 

It would make it easier to work with on the CMS, since we wouldn't have to update EDS anytime we want to use a different background color.

coleEthos (changes requested):
Looking at the example provided by design `https://www.ladderlife.com/`, we need to have a smoother transition here between the unscrolled and scrolled background colors. 

You should be able to implement either an `animation` + `keyFrames` in the CSS or define a `transition` associated with `background-color`.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
hangngo21: The color for nav bar is transparent by default and will only change to a different color when navbarColor is specified https://github.com/getethos/ethos-design-system/pull/492/files#diff-0218fe2c9ebf497571c90a9dc4cd522757d8d507b155bb70b1cabce02f3dad69R111
If the navbarColor is specified, then we will display the nav bar with the color passed by CMS. bgMoss class is the result of navbarColor='moss'.

hangngo21: From what we talk today, I will remove the navbarColor since the color of the nav bar will depend on the single CTA's background color and the changes are being worked on in CMS.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
```diff
@@ -89,6 +88,18 @@ $laptop-range-small-end: 929px;
...
+
+:global(.isScrolled) {
+  .laptopAndUpScrolled {
+    background-color: var(--White);
+    color: var(--GrayPrimary--translucent);
+  }
+}
```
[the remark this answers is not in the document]
hangngo21: Yes, I just added the 1s transition for the background color when user scrolling up and down

coleEthos:
@hangngo21 looking good. 

Can we add the fade-in-out transition for the bottom border as well? When the user is at the top of the screen without having scrolled and `animateNavbar` is `true`, the bottom border should be hidden.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
```diff
@@ -89,6 +88,24 @@ $laptop-range-small-end: 929px;
...
+
+:global(.isUnscrolled) {
+  .laptopAndUpAnimation {
+    transition: background-color 1s ease-in-out;
+  }
+}
+
+.laptopAndUpAnimation {
+  border-bottom: none;
+}
```
coleEthos, line 107: @hangngo21 it's a little unclear in the mocks, but I think the border should be animated, same as the background color, so that it fades in when `.isScrolled` is true and be hidden when `.isUnscrolled` is true.

coleEthos (approved):
LGTM, please make sure to add `v1.3.16` tag before merging.

src/components/UniversalNavbarExpanded/UniversalNavbarExpanded.module.scss
```diff
@@ -89,6 +89,26 @@ $laptop-range-small-end: 929px;
...
+
+:global(.isUnscrolled) {
+  .laptopAndUpAnimation {
+    transition: 1s ease-in-out;
+  }
+}
+
+.laptopAndUpAnimation {
+  border-bottom: none;
+  background-color: transparent;
```
coleEthos, line 109: minor: I think you can have the `transition: 1s ease-in-out;` rule in this `.laptopAndUpAnimation` just once instead of in two locations above. Not blocking.

hangngo21: Good point. I just added the changes.

PR #495, opened 2023-02-27
[GC-1964] [NSLP follow up] On navigate back, useEffect to set initial background to white if scrolled

hangngo21:
## [Jira Task](https://ethoslife.atlassian.net/browse/GC-1964)
This PR fix the issue when navigating back, the navigation bar should turn to white color if the user is not at the top of the page

https://github.com/getethos/cms/pull/7030

Screenshots and manual test instruction is in this PR https://github.com/getethos/cms/pull/7030

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethos-design-system",
-  "version": "1.3.19",
+  "version": "1.3.18",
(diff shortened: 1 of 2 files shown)


coleEthos (approved):
LGTM - reminder to push tag as well.

PR #499, opened 2023-05-11
Required changes to use in NextJS

arnas007:

-

### DIFF
--- a/src/components/Animation.scss
+++ b/src/components/Animation.scss
@@ -1,7 +1,7 @@
 /* @getethos/design-system/Animation.scss
    ========================================================================== */
 
-:root {
+.body {
   --transition-default: all 200ms cubic-bezier(0.21, 0.61, 0.35, 1);
   --transition-duration-short: 200ms;
   --transition-duration-medium: 500ms;
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -5,7 +5,7 @@
 
 /* Defaults */
 
-:root {
+.body {
   --ButtonMediumStatefulMinWidth: 156px;
 }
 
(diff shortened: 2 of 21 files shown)


src/components/Animation.scss
```diff
@@ -1,7 +1,7 @@
 /* @getethos/design-system/Animation.scss
    ========================================================================== */
 
-:root {
+.body {
```
coleEthos, line 4: I'm a little concerned with doing this in the main application and CMS, so we should test it first to make sure there are no regressions. Before we make this an official EDS version I think we need to do the following tests, with the package.json files pointed to this branch:

`"ethos-design-system": "github:getethos/ethos-design-system#arnas/GC-2299"` 

1. Open a CMS PR, find the <body> tag that needs this class, make sure there are no conflicts.
2. Open a main app PR for `/partners/bond/package.json` and `/frontend/package.json`. Again find the <body> tags (may be more than 2) and make sure there are not conflicts.

If you don't have access to main app `github.com/getethos/ethos`, can you submit an IT ticket request for read/write access there? We'll get you onboarded there next week.

As for the nextJS repo, I'm going to block this PR for now but we can use this EDS branch for our dev work.

coleEthos (changes requested):
Blocking until we can test CMS and main app for regressions. 

We should also post in #dev ( I will do this ) to callout these changes in case there are other consumers of EDS that we don't know about.

coleEthos:
@arnas007 can you please add some QA screenshots of the updated components? 

Once we have PRs open to test on CMS and main app, we can add them to the PR description here.

coleEthos:
Left a note here in #dev: https://getethos.slack.com/archives/CD0CEMEJ1/p1683868532671419

PR #500, opened 2023-05-23
[MAR-724] universnal nav bar redesign

darkul1:

# [Deploy Preview](https://deploy-preview-7366--ethos-cms.netlify.app/)

image

image


arnas007 (approved):

src/components/UniversalNavbarExpanded/CtaButton.module.scss
```diff
@@ -1,4 +1,5 @@
 @import '../Media/Media.scss';
+@import '../Variables.scss';
```
coleEthos, line 2: Why did this need to be added?

coleEthos (commented):
Overall code LGTM. Will wait until final design QA from Loc has been addressed and CMS PR has been updated with latest (both consumer and agents menu links) before approving. Nice work Darius!

coleEthos (changes requested):
Can we eliminate the shifting of the text when bolded, so it functions like the old bolding of text when hovered? After that I think this is good to go!

See in attached screen recording how the text shifts left to right when bolded upon hover, but does not move on staging.

https://github.com/getethos/ethos-design-system/assets/87672141/0a009820-6236-4fd3-8c46-00f651662906

darkul1:
> Can we eliminate the shifting of the text when bolded, so it functions like the old bolding of text when hovered? After that I think this is good to go!
> 
> See in attached screen recording how the text shifts left to right when bolded upon hover, but does not move on staging.
> 
>  Screen.Recording.2023-06-02.at.5.40.46.PM.mov

@coleEthos resolved this!

coleEthos (changes requested):
Could we make one more minor change here? Make the clickable area the button only, so you don't accidentally click this CTA when expanding/closing accordion menus on the right side of the screen? I did this once when testing on my phone. 

Screen Shot 2023-06-08 at 10 57 14 PM

coleEthos (approved):
LGTM, remember to add git tag with version number as described in readme, before or after merge!

PR #501, opened 2023-07-13
Ignas/cd 39

ignasstanionis:
## [Jira Task](https://ethoslife.atlassian.net/browse/CD-39)

Let's discuss in comments how it would be best to solve shared EDS typography.

EDS
-

image

As you can see typography is shared through CMS/Ethos monorepo (mostly only frontend parts)

The ticket idea is to increase small fonts per 1-2 pixels which for me seems not a big issue to increase it in EDS itself (at least for now). Caption becomes 16 instead of 14 and 12 becomes 14. If we want to increase it only on pre interview / interview part i can append 1 more styles in that case we won't decouple frontend from EDS, but it would take much more time to go through all the code and change it. Also i am not sure what is planned for EDS refresh yet. Is it better to wait for EDS updates and on hold this issue, or do 1 or another solution? Could you give your thoughts while you probably has more experience about how EDS shared through projects ? 
P.S. Some places decoupled from EDS and just uses style attribute.

@coleEthos @t1mwillis

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+@fortawesome:registry=https://npm.fontawesome.com/
+//npm.fontawesome.com/:_authToken=72A953AE-85E8-4159-B556-2EFDB780A2D7
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-@fortawesome:registry=https://npm.fontawesome.com/
-//npm.fontawesome.com/:_authToken=72A953AE-85E8-4159-B556-2EFDB780A2D7
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,0 @@
-nodeLinker: node-modules
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+nodeLinker: node-modules
--- a/src/components/TypeBase/TypeBase.module.scss
+++ b/src/components/TypeBase/TypeBase.module.scss
@@ -466,8 +466,8 @@ h6,
 
 .Eyebrow2 {
   &.Theinhardt {
-    $font-size: 14;
-    $line-height: 16;
+    $font-size: 12;
+    $line-height: 14;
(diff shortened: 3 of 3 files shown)


coleEthos (commented):
I think we need to have a discussion about how the new EDS work is going to be undertaken. I was under the impression that we'd be building a new standalone EDS, within the main app, for re-skinning PreInterview/Interview. There has been a long ongoing discussion between @t1mwillis @AjaySingh5 and myself about this project, and we decided on building a new EDS using Tailwind, rather than continuing to use the existing EDS. cc @michaelvmata 

I think this change is pretty risky (even though it's just a couple pixels) and has the potential to break a lot of component layouts across various consumers of EDS, such as main app and CMS. We'd need to open PRs there (deploy previews on CMS and release.com previews on Main App) and QA a lot of usages of Caption/Caption2, to make sure nothing is broken, before we merge this change into `master` on EDS.

For the purposes of this ticket, I think we should either: 

A) Move individual components with a small font size in PreInterview/Interview onto a larger existing typography style.

B) Initialize a new EDS in PreInterview, with these updated typography styles and then migrate individual components that need larger font size, to the new EDS.

Thoughts?

ignasstanionis:
I would say for now we can make A) fixing interview/PreInterview with existing EDS typo. And after that we can have a meeting to discuss about new EDS with tailwind and smth like storybook, of course design of most important components and variations would be great to have first, we could start from components which mostly used in projects and try migrate into new EDS bit by bit. Thanks for quick response and let me know about plans for new EDS. I will try to take care of interview fonts w/o braking current typography @coleEthos

coleEthos:
Thanks @ignasstanionis 

It sounds like maybe the new EDS and this improvement to PreInterview are separate initiatives. For this task I think it would be better to just leave the existing EDS caption component alone and build something bespoke for larger caption fonts in PreInterview. This might all be re-skinned with the new EDS work though, so I'm going to talk with @t1mwillis and @michaelvmata about what order it makes sense to do this work in.

PR #508, opened 2023-08-21
Ignas/cd 81 mobile

ignasstanionis:

-


coleEthos:
@ignasstanionis can you also update PR in cms-next to remove variables that have 24px vs 16px difference

ignasstanionis:
> @ignasstanionis can you also update PR in cms-next to remove variables that have 24px vs 16px difference

Done

PR #525, opened 2024-05-16
upgrade to yarn 4 and set fortawesome auth token

perryao:
## [Jira Task]()

blocks https://github.com/getethos/ethos/pull/29802

Trying to update the frontend to use yarn 4, but running into issues where yarn 4 reverts back to yarn 1 when building this package

- bump to yarn 4
- update to typescript 4
- set auth token for fortawesome
- bump node-sass to 9.0.0 because the version we were using didn't seem to support arm64 macs
- update webpack 5

-


tylermichael (approved):

coleEthos:
This LGTM once we confirm functionality in main app frontend.

coleEthos:
Can we add a URL to dev or release environment and get UAT from QA and product to help check for any visual or functional regression caused by upgrades?

.github/workflows/deploy.yml
```diff
@@ -24,13 +24,8 @@ jobs:
...
           # Don't share with other workflows because we are using `--production=false`
           key: dev-dependencies-${{ hashFiles('**/yarn.lock') }}
-      # We need to create a .npmrc with the fontawesome token
       - name: Setup fontawesome dependencies
-        run: |
-          cat > .npmrc <<EOF
-          @fortawesome:registry=https://npm.fontawesome.com/
-          //npm.fontawesome.com/:_authToken=${{ secrets.FONTAWESOME_TOKEN_GITHUB }}
-          EOF
+        run: yarn config set -H npmScopes.fortawesome.npmAuthToken ${{ secrets.FONTAWESOME_TOKEN_GITHUB }}
```
tylermichael, line 28: @coleEthos Do you know if we are allowed to rehost fontawesome in our own package registry? If so, we could do that and get rid of this extra token and simplify things further.

Considering you can serve it from your own server as [described here](https://docs.fontawesome.com/web/setup/host-yourself/svg-js), I feel like this might be within the TOS.

tylermichael: Could also remove the chance where we run out of bandwidth allowances or run into what happened recently where the charge failed.

coleEthos: @tylermichael it seems like we could likely do that without violating TOS, it seems to me like something that counts as self hosting, even if their documentation doesn't describe how to implement what you're proposing.

Want me to email and check with support to confirm?

coleEthos (approved):
lgtm for testing

.github/workflows/deploy.yml
tylermichael: Yea that might be best just to be safe. I'd be surprised if that goes against TOS.

I would mention that it would be hosted in a private NPM registry that requires an auth key etc.

coleEthos: Okay emailing them now.

coleEthos: @tylermichael I cc'd you but this was their response: 
> As long as you are not providing the pro assets to non-seat users you should be good.

PR #526, opened 2024-05-16
full story mask for DoB

coleEthos:
## [Jira Task](https://ethoslife.atlassian.net/browse/GC-4008)
Adds the fullstory exclusion mask as an optional parameter on the DateInput component that is used by date of birth (DoB) frontend input on the monorepo:

`BirthdateInput` from this EDS package in main app is only imported here: 
```
frontend/src/main/InsuranceApp/PreInterview/Questions/components.tsx
```

DateInput remains backwards compatible as the param is optional. 

Tested in main app in this PR:

- See link above.


src/components/DateInput/DateInput.js
```diff
@@ -84,7 +85,7 @@ const PrivateDateInput = (props) => {
   })
 
   const getClasses = () => {
-    const base = `DateInput ${styles.TextInputCommon}`
+    const base = fullstoryMask ? `DateInput ${styles.TextInputCommon} fs-exclude` : `DateInput ${styles.TextInputCommon}`
```
coleEthos, line 88: https://help.fullstory.com/hc/en-us/articles/360020623574-How-do-I-protect-my-users-privacy-in-Fullstory#01F5DPW1AJMP73R81PE54AENFE

siddrajpoot (approved):

